PR #60, opened 2023-04-05
Support string input for SOR public functions

danielmkm:


danielmkm:
No worries at all man, it turned out to be kinda hacky in the end

PR #65, opened 2023-05-16
In instances where the index is not present in the subgraph, we default to the array index

danielmkm:


src/data/providers/subgraphPoolProvider.ts
```diff
@@ -166,6 +166,16 @@ export class SubgraphPoolProvider implements PoolDataProvider {
             pools = pools.filter((pool) => this.poolMatchesFilter(pool));
         }
 
+        // Until a full graft finishes on polygon/arbitrum, there are instances where the index is not
+        // present. For these instances, we default to the index of the token in the array.
```
mikemcdonald, line 170: Is it deterministic that subgraph returns the pool tokens in the order expected here for all pool types?

PR #68, opened 2023-06-13
Add Gyro2 pool

brunoguerios:
Main:
- Add Gyro2 pool based on SOR implementation.

Other changes:
- add .nvmrc to make it easier to keep track of which node version we should use within the package
- update pnpm-lock.yaml (which as done automatically by running `pnpm install` - not sure if I should commit this - please let me know if I should revert that 🙏)

Ps: I followed what was stablished by other pools and didn't implement `getNormalizedLiquidity` - I'd say we should be able to add that on a following PR in case it's needed.
Ps2: feedback and improvement suggestions are more than welcome - I find PR reviews one of the best way to learn 😊


src/entities/pools/gyro2/gyro2Pool.ts
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    public swapGivenIn(
+        tokenIn: Token,
+        tokenOut: Token,
+        swapAmount: TokenAmount,
+        mutateBalances?: boolean,
+    ): TokenAmount {
+        // TODO: try/catch is being used on previous SOR - should we keep the same bevaiour or should we throw an error?
+        try {
```
mikemcdonald, line 122: No blanket try / catches - we should understand any failure scenarios and have detailed throws

src/entities/pools/gyro2/gyro2Pool.ts
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    derivativeSpotPriceAfterSwapGivenOut(
+        tokenIn: Token,
+        tokenOut: Token,
+        swapAmount: TokenAmount,
+    ): bigint {
+        console.log(tokenIn, tokenOut, swapAmount);
+        throw new Error('Not implemented');
+    }
```
mikemcdonald, line 270: Remove these. This previous concept of swap amount optimization isn't used in `b-sdk`

mikemcdonald (commented):
Left a few small comments. For test's imo we should keep everything in `bigints`. Seeing things like `Number()` is a code smell - any time numbers are represented in decimal formats throughout our codebases should only be for display purposes as strings.

.nvmrc
```diff
@@ -0,0 +1 @@
+18.16.0
```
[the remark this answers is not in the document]
mikemcdonald: We can recommend users using v16 use a polyfill for fetch. Similar to https://viem.sh/docs/compatibility.html#polyfills

brunoguerios:
> Left a few small comments. For test's imo we should keep everything in bigints. Seeing things like Number() is a code smell - any time numbers are represented in decimal formats throughout our codebases should only be for display purposes as strings.

I researched a bit and found out the ability to add custom jest matchers, which should allow us to compare `bigint` values on assertions.
I still couldn't make it work 100%, but I got to a point where I spent too much time on it to give up 😬 😅

I decided to push changes anyway hoping someone has experience with that and is able to help out 🙏 
Basically, if I include `jest.d.ts` on `tsconfig`, tests work as expected, but for some reason linter warns that the method doesn't exist. If I remove `jest.d.ts` from `tsconfig` linter finds the method, but tests fail stating the method doesn't exist 🤦

brunoguerios:
Never mind - @johngrantuk suggested using ChatGPT and it was able to figure out the issue in minutes. My implementation was not compatible with the latest jest version. Now it's fixed ✅

mikemcdonald (approved):
Tests look great - will try to incorporate some of this into other pool types. I'll tag a release once I get a few other small updates in

PR #71, opened 2023-06-19
Add GyroEV2 pool

brunoguerios:
- Add GyroEV2 pool by extending GyroE pools to handle token rates

Note:
- this implementation defaults GyroE v1 pools to have token rates of 1 and reuses same pool logic
- this PR adds integration tests that run on Polygon, so it needs `POLYGON_RPC_URL` as env var (and it should also be added to github secrets in order to have PR checks working as expected)


johngrantuk (commented):
I think the onChainPoolDataEnricher needs to be updated to get the GyroE V2 tokenRates?

The current poolDataQuery helper doesn't support that. Could be done using a separate mulitcall like [this](https://github.com/balancer/balancer-sdk/pull/476/files#diff-62b9b89d378ae44c343c6244055d2b1aea5bb3432aa285aa3eb5c40cdd815f04). It would be more convenient if we updated the poolDataQuery helper but not sure if that's a reasonable approach?

PR #74, opened 2023-06-21
Add Gyro3 pool

brunoguerios:
Main
- Add Gyro3 pool

Notes
- We're missing unit tests for gyro3 math (which is not available in the SOR). I'd say we're ok because we have unit tests for gyro3 pool methods, which rely on gyro3 math, but let me know if you think we still should have unit tests covering those.


src/entities/pools/gyro2/gyro2Pool.ts
```diff
@@ -247,8 +247,8 @@ export class Gyro2Pool implements BasePool {
         sqrtAlpha: bigint;
         sqrtBeta: bigint;
     } {
-        const tIn = this.tokenMap.get(tokenIn.wrapped);
-        const tOut = this.tokenMap.get(tokenOut.wrapped);
+        const tIn = this.tokenMap.get(tokenIn.address);
```
mikemcdonald, line 250: Any reason to update this logic?

src/entities/pools/gyro3/gyro3Math.ts
```diff
@@ -0,0 +1,254 @@
+import {
+    _INVARIANT_MIN_ITERATIONS,
+    _INVARIANT_SHRINKING_FACTOR_PER_STEP,
+} from './constants';
+import { _safeLargePow3ADown } from './helpers';
+import { MathGyro, ONE } from '../../../utils/gyroHelpers/math';
```
mikemcdonald, line 6: Switch any `ONE` consts to use the global `WAD`.`ONE` can be confusing because it's still not clear which scale is being used

src/entities/pools/gyro2/gyro2Pool.ts
brunoguerios: I considered this an oversight from my side. I copied code from other pools and didn't change that.
I'm not really sure what this `wrapped` means, so I thought it would make more sense to get the token based on its `address`.
Should I revert it to use `wrapped` instead?

mikemcdonald: `wrapped` allows the sor and pool side of things to use the same `Token` class - so that native ETH (and similar assets) can be passed around while any pool / math uses the `wrapped` balance. If a `Token` class is not initialized with a different `wrapped` address it defaults to the token address.

I'll think through if there's a cleaner way to do this now that repo structure is mostly in place

brunoguerios: Got it! I reverted to use wrapped instead of address for now. ✅

mikemcdonald (approved):

PR #84, opened 2023-07-14
Fix Token Amount with zero decimals

brunoguerios:


brunoguerios:
Closing this as decimals zero is not actually a problem here. 
The real problem is actually related to this discussion: https://github.com/wagmi-dev/viem/discussions/490

PR #90, opened 2023-08-23
drop in replacement for pool data via multicall

gmbronco:
Adding multicall source for getting pool's on chain data. It's meant as a drop-in replacement for the current queries contract. Would appreciate pointers where the pools query was failing so we can double check if multicall works as expected.

### DIFF
--- a/debug/multicall.ts
+++ b/debug/multicall.ts
@@ -1,16 +1,21 @@
-import { fetchAdditionalPoolData } from '../src/data/onChainPoolDataViaMulticall';
 import { OnChainPoolDataViaMulticallEnricher } from '../src/data/enrichers/onChainPoolDataViaMulticallEnricher';
 import { createPublicClient, http } from 'viem';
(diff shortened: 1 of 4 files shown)


johngrantuk:
@gmbronco I made the changes to allow us to add Fantom config in [this PR](https://github.com/balancer/b-sdk/pull/69/files) and came across the issues: 

getPoolData call would randomly fail. Seems to be related to number of pools although the error message changed over a certain amount. I saw the following result consistently:
* Up to 370 pools - ok
* 371 pools - Error: "RPC Request failed, execution reverted"
* 380 pools - Error: "out of gas"
I tried a few different RPC endpoints and all seemed the same.
Also to rule out a specific pool at index>370 causing issues I tried from index: 300->End and it ran ok.
One day it failed almost every time but another it seems to be much less.

Also there is an ongoing issue on zkEVM so would be interesting to see if this works there too.

gmbronco:
how do i reproduce, do you use any specific subgraph query and chain?

PR #91, opened 2023-08-29
Add Add/Remove Liquidity

brunoguerios:
- Add add/remove liquidity support (non-nested pools)
- Uses balancerHelpers to query amounts in/out rather than relying on specific pool math and associated data
- Tests against local Anvil (also CI update)
  - Weighted Pool
  - CS Pool
  - Gyro Pool
- API Provider added
- Example added for add/remove liquidity


src/entities/join.ts
```diff
@@ -0,0 +1,25 @@
...
+    constructor(config?: JoinConfig) {
+        const { customPoolFactories } = config || {};
+        this.poolFactories = {
+            weighted: new JoinWeighted(),
+            // custom pool factories take precedence over base factories
+            ...customPoolFactories,
+        };
+    }
+
+    public getJoin(poolType: string): BaseJoin {
```
johngrantuk, line 21: `getJoin` seems a bit confusing as we're not really getting the join info? What about something like `getHelper` or changing class to `PoolJoins` and function to `getJoinHelper`?

src/entities/pools/weighted/weightedJoin.ts
```diff
@@ -0,0 +1,187 @@
+import { BaseJoin, JoinInput, JoinQueryResult, PoolState } from '..';
```
johngrantuk, line 1: Looks good so far I think. Probably just need to keep in mind when building it out how we can keep it modular so we can reuse when it comes to chaining as part of a relayer call.

src/entities/pools/index.ts
```diff
@@ -33,3 +33,39 @@ export interface BasePoolFactory {
     isPoolForFactory(pool: RawPool): boolean;
     create(chainId: number, pool: RawPool): BasePool;
 }
+
+// Returned from API and used as input
+export type PoolState = {
+    id: string;
```
johngrantuk, line 39: We will likely move away from poolId in the future. So I wonder if it makes sense/is practical to start using pool address instead? That way we can keep a single code base or at least a consistent interface.

src/entities/pools/weighted/weightedJoin.ts
brunoguerios: I'll try to keep that in mind, but I feel like we'll be able to better identify what needs to be adapted only when we add chaining/relayer calls.

src/entities/pools/index.ts
brunoguerios: I thought about that, but `queryJoin` requires `poolId`, so we'll need it for now.
What we could do is add `address` to PoolState and use `id` only when it's strictly necessary in favor or `address` everywhere else.
I'll go ahead with that change for now, but feel free to suggest a different approach.

src/entities/pools/index.ts
```diff
@@ -33,3 +33,39 @@ export interface BasePoolFactory {
...
+    assets: string[];
+    joinKind: string;
+    bptOut: TokenAmount;
+    amountsIn: TokenAmount[];
+};
+
+export interface BaseJoin {
+    getInstance(): BaseJoin;
+    query(input: JoinInput, poolState: PoolState): Promise<JoinQueryResult>;
+    // TODO - Best way to represent slippage?
```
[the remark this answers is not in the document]
johngrantuk: My only concern about that is that more options could lead to more mistakes. Something to ask Daniel I think.

src/entities/pools/index.ts
johngrantuk: At this stage that sounds good. I can't remember how poolId is actually calculated but I thought there might be a helper to do address>poolId. Another thing to check with Daniel to see what he thinks about that kind of direction in general.

src/entities/pools/index.ts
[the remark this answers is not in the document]
brunoguerios: I think that the main benefit is that it decouples the unit provided from the unit used within the code.
The user is able to provide whatever unit they already have without having to transform on their end.
Plus it's easier to understand all possible input formats by checking the object interface.
At least that has been my experience with TokenAmount

src/entities/pools/index.ts
brunoguerios: We can always query `getPoolId` onChain, but if we do go down that path, I feel like it makes more sense to do it on the API level.

src/entities/pools/weighted/weightedJoin.ts
```diff
@@ -0,0 +1,226 @@
...
+        } else if (tokensIn.includes(poolState.address)) {
+            if (tokensIn.length > 1) {
+                throw new Error('Cannot join with BPT and other tokens');
+            } else if (input.isInit) {
+                throw new Error('Cannot init with BPT');
+            }
+        }
+    }
+
+    private getJoinKind(input: JoinInput, poolAddress: Address): string {
```
danielmkm, line 205: Would type the return value here better, string union maybe.

src/entities/pools/weighted/weightedJoin.ts
```diff
@@ -0,0 +1,226 @@
...
+        let userData: Address;
+        const poolAssets = poolState.tokens.map((t) => t.address);
+
+        switch (joinKind) {
+            case 'Init': {
+                maxAmountsIn = this.getAmountsIn(input, poolAssets);
+                userData = WeightedEncoder.joinInit(maxAmountsIn);
+                break;
+            }
+            case 'GivenIn': {
```
danielmkm, line 50: imo we should take this opportunity as a chance to get away from these hard to reason about name. I would lean more towards mapping to things like `proportional`, `unbalanced`, `singleAsset`, etc

src/entities/pools/weighted/weightedJoin.ts
```diff
@@ -0,0 +1,226 @@
...
+import { vaultAbi } from '../../../abi';
+
+export class WeightedJoin implements BaseJoin {
+    // TODO - Probably not needed
+    getInstance(): WeightedJoin {
+        return new WeightedJoin();
+    }
+
+    public async query(
+        input: JoinInput,
```
danielmkm, line 29: I get the goal here, but in trying to make a one size fits all, it seems like you're creating an interface that is quite challenging to reason about. For given out, I need to know that the `input.tokenAmounts.length === 1` and that it should only contain the bpt. The only way I would be able to figure that out is by reading the docs or reading the code, I don't think I would ever intuitively know that.

src/entities/pools/weighted/weightedJoin.ts
```diff
@@ -0,0 +1,226 @@
...
+            poolId: poolState.id,
+            assets: poolAssets,
+            sender: ZERO_ADDRESS,
+            recipient: ZERO_ADDRESS,
+            maxAmountsIn,
+            userData,
+        });
+
+        const client = createPublicClient({
+            transport: http(input.rpcUrl),
```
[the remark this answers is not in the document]
jxom: We don’t perform any network requests on client init.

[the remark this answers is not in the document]
brunoguerios: By checking this is not the case with `viem` you mean that we should omit this input?
I'll make it optional in case someone wants the easy path of using default values, but we also allow for custom rpcUrls just in case. 
Let me know if you think it's better to avoid that complexity at all.

src/entities/pools/weighted/weightedJoin.ts
brunoguerios: I think that a good approach to fix this would be John's initial draft, where he provided different input types, that make it easier for the user to make sense of all possible join variants.

Since it's a small change, I'll go ahead and update the PR, so we can follow-up the discussion with more concrete implementation.

src/entities/pools/weighted/weightedJoin.ts
brunoguerios: I get your point of renaming to make the interface more intuitive. That was the goal of replacing `ExactTokensInForBPTOut` by `GivenIn` 😅 but I agree that this can be improved.

Just trying to understand what you mean by differentiating `proportional` and `unbalanced` though. Is that in the sense of having some sort of helper that would automatically calculate proportional amounts in based on a single asset amount in as reference? I can see that being helpful, but just wanted to make sure this was the suggestion.

And I'd love some help on naming the `givenOut` alternative - maybe `exactOut`?

danielmkm: I'd probably do something like `SingleAssetExactAmountIn` `SingleAssetExactBptOut`

src/entities/join/index.ts
```diff
@@ -0,0 +1,86 @@
...
+// Returned from API and used as input
+export type PoolState = {
+    id: Address;
+    address: Address;
+    type: string;
+    // TODO: is it ok to replace by Token here? Or should we stick to basic types?
+    tokens: {
+        address: Address;
+        decimals: number;
+    }[]; // already properly sorted in case different versions sort them differently
```
danielmkm, line 22: Might want to include the `index` here and not rely on the tokens already being sorted as it creates a hidden requirement.

src/entities/join/weighted/weightedJoin.ts
```diff
@@ -0,0 +1,149 @@
...
+} from '..';
+
+export class WeightedJoin implements BaseJoin {
+    // TODO - Probably not needed
+    getInstance(): WeightedJoin {
+        return new WeightedJoin();
+    }
+
+    public async query(
+        input: JoinInput,
```
danielmkm, line 34: would generally maybe find a more explicit way to deal with native asset joins. Perhaps a flag on `BaseJoinInput` called something like `joinWithNativeAsset?: boolean` and not expecting a zero address, as it is another hidden requirement that is easy to overlook.

src/entities/join/weighted/helpers.ts
```diff
@@ -0,0 +1,135 @@
...
+        case JoinKind.SingleAsset:
+            if (input.amountIn.token.address === ZERO_ADDRESS)
+                nativeTokenIn = input.amountIn;
+            break;
+    }
+    let assets = [...poolAssets];
+
+    if (nativeTokenIn !== undefined) {
+        const wrappedAssetIndex = poolAssets.findIndex(
+            // TODO: this assumes that the user provided wrapped address for input tokens - is that a fair assumption?
```
danielmkm, line 125: This is another tough one, as it is creating layers of requirements that are very easy to forget or just not even know about.

src/entities/join/index.ts
johngrantuk: I think here we were working on assumption that it will be leveraging the API to provide information in correct format, etc but I guess we're still working out limits.

src/entities/join/weighted/weightedJoin.ts
```diff
@@ -0,0 +1,149 @@
...
+        poolState: PoolState,
+    ): Promise<JoinQueryResult> {
+        // TODO - This would need extended to work with relayer
+
+        checkInputs(input, poolState);
+
+        const poolAssets = poolState.tokens.map((t) => t.address);
+        const assets = getAssets(input, poolAssets);
```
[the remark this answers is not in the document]
brunoguerios: you mean the `getAssets` method? it's there to make sure we use ZERO_ADDRESS when needed
if we do go down the path of having a `joinWithNativeAsset`, this will likely not be needed anymore

src/entities/join/weighted/weightedJoin.ts
brunoguerios: yeah, I'm so used to think of native asset joins with zero address that I didn't think of that - I agree it's a hidden requirement that we can make more explicit.
I'm assuming that with that approach we would expect the `wrapped` asset as input and replace that with `zero address` if `joinWithNativeAsset = true`, right?
Would we also accept `zero address` as input? Or should we throw if the user tries that?
I ask because limiting the input to pool assets only and not having to handle the case where `zero address` is passed would reduce a lot the complexity there.

src/entities/join/weighted/helpers.ts
brunoguerios: yeah, I think we should not expect the user to provide that info - if we decided to limit the user input to pool tokens only, handling this native/wrapped addresses won't be needed anymore, so we should be ok
if we do accept `zero address` as input, we'll likely need a native/wrapped mapping hardcoded within b-sdk (as we currently have on the SDK)

src/entities/pools/weighted/weightedJoin.ts
brunoguerios: Regarding naming, I was trying to move away from being that verbose, but I’m having a hard time thinking about alternatives.
I did find those old `ExactTokensInForTokenOut` pretty confusing, but thinking a bit more about it I’d say the confusion came from having `TokenIn` and `TokenOut` in the same name. I’d always have to read a few times to make sure I was using the correct function.
Having said that, when I read `SingleAssetJoin` and `ExactBptOut` in the same name, I kinda expect it to perform a single asset join, but with an exact amount out. Which is not the case, right? I's actually performing a `proportional` join with an exact amount out.
In any case, I'd still consider a `ProportionalExactAmountsInJoinInput` + `ProportionalExactBptOutJoinInput` not as good as simply `ProportionalJoinInput` + `ExactBptOutJoinInput`.
In fact, I'd actually go for `ExactAmounOutJoinInput` - which seems easier to understand than `Bpt` for devs that are not super into Balancer stuff.
Let me know what you think

.github/workflows/checks.yml
```diff
@@ -42,6 +42,8 @@ jobs:
       - uses: actions/checkout@v3
       - name: Setup
         uses: ./.github/actions/setup
+      - name: Run local fork in background for integration tests
+        run: npx hardhat --tsconfig tsconfig.testing.json node --hostname 127.0.0.1 --fork ${{ secrets.ETHEREUM_RPC_URL }} &
```
johngrantuk, line 46: Not a priority now but worth considering moving to [Anvil](https://book.getfoundry.sh/anvil/#how-to-use-anvil) instead of Hardhat? I think its a good bit faster

src/entities/join/weighted/weightedJoin.ts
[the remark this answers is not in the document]
brunoguerios: Oh, I get what you mean. Since it's a query, it doesn't matter if we're doing with native/wrapped address.
I do use the result of that method to build `JoinQueryResult` that `query` returns (which is then relevant to build the transaction itself) but as I said, this will all likely change with the new approach.

src/entities/pools/weighted/weightedJoin.ts
brunoguerios: After talking to Daniel, I realized that my understanding of the request of moving away from current names was different from what he was expecting.
I thought the idea was to rethink the whole interface towards the user interaction and then map those options down to the contract interface.
For example, we'd have `proportional` + `unbalanced` + `singleAsset` on the SDK interface, with different inputs and logic, but they would all map to down to `EXACT_TOKENS_IN_FOR_BPT_OUT` in the contract interface.
That was not his intention. He wanted to simply rename the 4 options we currently have on the contract interface to something easier to read/remember and also more descriptive. For example:
- `INIT` -> `Init`
- `EXACT_TOKENS_IN_FOR_BPT_OUT` -> `Unbalanced`
- `TOKEN_IN_FOR_EXACT_BPT_OUT` -> `SingleAsset`
- `ALL_TOKENS_IN_FOR_EXACT_BPT_OUT` -> `Proportional`

.github/workflows/checks.yml
brunoguerios: I moved this from a subtask to a task on asana, so it doesn't block the PR

src/entities/join/weighted/helpers.ts
```diff
@@ -0,0 +1,64 @@
...
+    sender: Address;
+    recipient: Address;
+    maxAmountsIn: readonly bigint[];
+    userData: Address;
+}) {
+    const joinPoolRequest = {
+        assets, // with BPT
+        maxAmountsIn, // with BPT
+        userData, // wihtout BPT
+        fromInternalBalance: false,
```
johngrantuk, line 23: I think we should make this part of the input parameters so the user can choose?

src/entities/join/weighted/weightedJoin.ts
```diff
@@ -0,0 +1,208 @@
...
+        minBptOut: bigint;
+        maxAmountsIn: bigint[];
+    } {
+        let maxAmountsIn = input.amountsIn.map((a) => a.amount);
+        let minBptOut = input.bptOut.amount;
+        let userData: Address;
+
+        switch (input.joinKind) {
+            case JoinKind.Init: {
+                maxAmountsIn = input.amountsIn.map((a) => a.amount);
```
johngrantuk, line 142: I find it a lot easier to reason about what `maxAmountsIn` and `minBptOut` should be for each switch case when they are both defined within the case rather than relying on previously defined default. (See [Exit](https://github.com/balancer/b-sdk/blob/exit/src/entities/exit/weighted/weightedExit.ts#L38))

src/entities/join/parser.ts
```diff
@@ -0,0 +1,25 @@
...
+        const { customPoolJoins } = config || {};
+        this.poolJoins = {
+            Weighted: new WeightedJoin(),
+            // custom pool Joins take precedence over base Joins
+            ...customPoolJoins,
+        };
+    }
+
+    public getJoin(poolType: string): BaseJoin {
+        // TODO - Need to parse
```
johngrantuk, line 22: Not sure what this TODO was referencing?

src/entities/join/weighted/weightedJoin.ts
```diff
@@ -0,0 +1,208 @@
...
+
+export class WeightedJoin implements BaseJoin {
+    public async query(
+        input: JoinInput,
+        poolState: PoolState,
+    ): Promise<JoinQueryResult> {
+        // TODO - This would need extended to work with relayer
+
+        // TODO: Extend input validation for cases we'd like to check
+        checkInputs(input, poolState);
```
johngrantuk, line 31: Maybe `validateInputs` is a better name here?

src/entities/join/index.ts
```diff
@@ -0,0 +1,82 @@
...
+export type JoinCallInput = JoinQueryResult & {
+    slippage: Slippage;
+    sender: Address;
+    recipient: Address;
+};
+
+export interface BaseJoin {
+    query(input: JoinInput, poolState: PoolState): Promise<JoinQueryResult>;
+    buildCall(input: JoinCallInput): {
+        call: Address;
```
johngrantuk, line 76: I think using `Address` for non-addresses is maybe a bit confusing. We should be able to create a new type easily enough.

src/entities/join/weighted/weightedJoin.ts
```diff
@@ -0,0 +1,208 @@
...
+} from '..';
+
+export class WeightedJoin implements BaseJoin {
+    public async query(
+        input: JoinInput,
+        poolState: PoolState,
+    ): Promise<JoinQueryResult> {
+        // TODO - This would need extended to work with relayer
+
+        // TODO: Extend input validation for cases we'd like to check
```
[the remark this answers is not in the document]
brunoguerios: I added the comment intentionally.
I didn't give much thought into all input validation we should make on the SDK level and this TODO is a reminder to revisit that at some point.
On the other hand, I did create an Asana task to make sure we keep track of that, so I'd say we should be ok on removing the TODO 👍

src/entities/join/parser.ts
brunoguerios: This was left from the initial draft and I forgot to remove. Will do it now 👍

src/entities/join/index.ts
brunoguerios: You're right. We already have a `Hex` type that should be a better fit on those cases. I'll update them here.

src/entities/join/parser.ts
```diff
@@ -0,0 +1,24 @@
...
+    customPoolJoins: Record<string, BaseJoin>;
+};
+
+export class JoinParser {
+    private readonly poolJoins: Record<string, BaseJoin> = {};
+
+    constructor(config?: JoinConfig) {
+        const { customPoolJoins } = config || {};
+        this.poolJoins = {
+            Weighted: new WeightedJoin(),
```
danielmkm, line 15: This pattern needs revisiting, although doesn't need to be done in this PR. How do I as a consumer of the SDK know that the weighted pool is keyed on `Weighted`?

danielmkm, line 15: This pattern needs revisiting, although doesn't need to be done in this PR. How do I as a consumer of the SDK know that the weighted pool is keyed on `Weighted`?

src/entities/join/parser.ts
```diff
@@ -0,0 +1,24 @@
...
+/*********************** Basic Helper to get join class from pool type *************/
+export type JoinConfig = {
+    customPoolJoins: Record<string, BaseJoin>;
+};
+
+export class JoinParser {
+    private readonly poolJoins: Record<string, BaseJoin> = {};
+
+    constructor(config?: JoinConfig) {
+        const { customPoolJoins } = config || {};
```
danielmkm, line 13: Also here, the idea behind supporting `customPoolJoins` is that I as the developer benefit somehow from using the SDK. That goal is not achieved with the current abstraction. I would still need to implement all functionality on my own.

If we take the SOR `customPoolFactories` as an example, by implementing a custom pool factory my pools are then included in the SOR routing.

brunoguerios: Looking at the weightedPool implementation, it feels like the only part that is poolType specific is the `userData` and everything else could be reused.
In any case, I'd say it will be easier to identify these opportunities once we implement a second join type. Then we rearrange methods to expect `customPoolJoins` to implement only what's necessary.
What do you think?

src/entities/join/weighted/weightedJoin.ts
johngrantuk: (Updated in refactor)

brunoguerios:
@johngrantuk - should I go ahead and merge this PR? Is there anything pending?

johngrantuk:
> @johngrantuk - should I go ahead and merge this PR? Is there anything pending?

Lets wait until we have the final interface definitions agreed then update to use those, e.g. addLiquidity vs join. That way we're not starting off with incorrect versions?

test/composableStableJoin.integration.test.ts
```diff
@@ -0,0 +1,276 @@
...
+            input = {
+                chainId,
+                rpcUrl,
+                kind: JoinKind.Unbalanced,
+            };
+        });
+        test('token inputs', async () => {
+            const joinInput = {
+                ...input,
+                amountsIn: [...amountsIn.splice(0, 1)],
```
lgahdl, line 114: this could be replaced by `amountsIn: amountsIn.slice(0,1)`, I think it's more clear

test/weightedJoin.integration.test.ts
```diff
@@ -0,0 +1,277 @@
...
+            input = {
+                chainId,
+                rpcUrl,
+                kind: JoinKind.Unbalanced,
+            };
+        });
+        test('with tokens', async () => {
+            const joinInput = {
+                ...input,
+                amountsIn: [...amountsIn.splice(0, 1)],
```
lgahdl, line 114: same here, could be `amountsIn: amountsIn.slice(0,1)`

src/entities/join/weighted/weightedJoin.ts
```diff
@@ -0,0 +1,187 @@
...
+        input: JoinInput,
+        poolState: PoolState,
+    ): Promise<WeightedJoinQueryResult> {
+        const amounts = this.getAmountsQuery(poolState.tokens, input);
+
+        const userData = this.encodeUserData(input.kind, amounts);
+
+        const { args, tokensIn } = parseJoinArgs({
+            useNativeAssetAsWrappedAmountIn:
+                !!input.useNativeAssetAsWrappedAmountIn,
```
lgahdl, line 30: @johngrantuk JoinInput type doesn't have this property "useNativeAssetAsWrappedAmountIn", should add this as a to do?

PR #92, opened 2023-09-04
fix: Add missing gyro3 to SupportedRawPoolTypes.

johngrantuk:


brunoguerios (approved):
I'm struggling to identify where this `SupportedRawPoolTypes` is being used.
Are we limiting the code in any aspect based on those supported types?
I ask that because this didn't show up on my tests. In other words, it seems to not have any impact on the code 🤔 
Am I missing something?

johngrantuk:
> 

I'm actually using it externally in the API [here](https://github.com/beethovenxfi/beethovenx-backend/blob/sor-v2/modules/sor/sorV2/sorV2.service.ts#L425). I assume its the best option but if there's an alternative maybe we can remove `SupportedRawPoolTypes`.

PR #93, opened 2023-09-13
Exit Pool

johngrantuk:


src/entities/common/index.ts
brunoguerios, line 1: Since it's common, why not simply have these files on the root of `entities` folder?

src/entities/exit/weighted/weightedExit.ts
```diff
@@ -0,0 +1,181 @@
...
+                                ?.amount ?? 0n,
+                    ),
+                    tokenOutIndex: undefined,
+                    maxBptAmountIn: MAX_UINT256,
+                };
+            case ExitKind.SINGLE_ASSET:
+                return {
+                    minAmountsOut: Array(tokens.length).fill(0n),
+                    tokenOutIndex: tokens.findIndex(
+                        (t) => t.address === input.tokenOut.toLowerCase(),
```
danielmkm, line 83: Would try to move sanitizing of input outside of the individual implementations. This isn't something we want to remember to do at this level, imo.

danielmkm (commented):
Overall looks really good. Clean and easy to read.

src/entities/exit/weighted/weightedExit.ts
johngrantuk: Thanks for the pointer, added `isSameAddress` to Token implementation.

danielmkm: This is fine, but it'd be nicer if sanitize happened before it ever got to `weightedExit`, so we don't even need to worry about this.

danielmkm:
Did a small refactor to show how we can create an intermediate layer to move some of the duplicate logic that will be in all exit types to a higher level. #96

PR #94, opened 2023-09-14
Refactor common pools

johngrantuk:
Separated out logic that will be common to other joins. Aim is to make it easier and safer for others to add new pools by reusuing as much as possible.
Also tried to simplify the logic for setting amounts (for query/call) and doing encoding.


src/entities/join/index.ts
```diff
@@ -9,28 +10,16 @@ export enum JoinKind {
...
-        decimals: number;
-        index: number;
-    }[];
-};
-
 // This will be extended for each pools specific input requirements
 export type BaseJoinInput = {
     chainId: number;
-    rpcUrl?: string;
+    rpcUrl: string;
```
brunoguerios, line 16: I made `rpcUrl` optional because viem provides a default value in case the user wants to use it.
I'm not against making it required, but I'd say that it doesn't have to be.

src/entities/types.ts
```diff
@@ -0,0 +1,21 @@
...
+export type TokenApi = {
+    address: Address;
+    decimals: number;
+    index: number;
+};
+
+export type Amounts = {
+    maxAmountsIn: bigint[];
+    tokenInIndex: number | undefined;
+    minimumBpt: bigint;
```
brunoguerios, line 20: It called my attention because I feel like I'm more used to `minBptOut`, but not a big issue, fine if we keep as is.

src/entities/join/weighted/helpers.ts
brunoguerios, line 1: Not sure we'll need other helpers in here, so might be interesting to replicate the pattern you set on `utils` and rename to `inputValidation.ts` or something similar.

brunoguerios (approved):
Thanks for the refactor! It looks way better 👏

src/entities/types.ts
johngrantuk: I went with `minimumBpt` because it matches the encoder parameter and thought it might be best to try and follow that as much as possible this time around.

src/entities/join/index.ts
```diff
@@ -58,7 +47,7 @@ export type JoinInput =
 
 // Returned from a join query
 export type JoinQueryResult = {
-    id: Hex;
+    poolId: Address;
```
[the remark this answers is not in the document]
johngrantuk: No, I must have missed it when I was merging the main branch. `Hex` | `Address` seems like a possibility although it doesn't really add any real type safety as its just checking for a 0x string.

src/entities/join/index.ts
johngrantuk: Good point. I've definitely had issues where default RPCs fail and its confusing so making it explicit to provide one might help to make that clearer (but maybe thats because original SDK is doing intensive multicalls). Could make it required for now and fine to change if we feel we need to.

src/entities/types.ts
```diff
@@ -0,0 +1,21 @@
...
+
+// Returned from API and used as input
+export type PoolState = {
+    id: Hex;
+    address: Address;
+    type: string;
+    tokens: TokenApi[];
+};
+
+export type TokenApi = {
```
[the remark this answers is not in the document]
johngrantuk: Yeah it was! Made a change here, let me know if it still doesn't feel right.

src/entities/join/index.ts
```diff
@@ -9,28 +10,16 @@ export enum JoinKind {
...
 // This will be extended for each pools specific input requirements
 export type BaseJoinInput = {
     chainId: number;
-    rpcUrl?: string;
+    rpcUrl: string;
     joinWithNativeAsset?: boolean;
     fromInternalBalance?: boolean;
 };
 
 export type InitJoinInput = BaseJoinInput & {
```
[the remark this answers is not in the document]
johngrantuk: I'll add an Asana ticket so it can be done in a separate PR.

src/entities/join/weighted/helpers.ts
```diff
@@ -1,42 +1,10 @@
...
-        assets, // with BPT
-        maxAmountsIn, // with BPT
-        userData, // wihtout BPT
-        fromInternalBalance,
-    };
-
-    return [poolId, sender, recipient, joinPoolRequest] as const;
-}
-
 export function validateInputs(input: JoinInput, poolState: PoolState) {
```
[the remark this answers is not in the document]
johngrantuk: Made a small change here anyway just to try and make it easier to follow.

brunoguerios (approved):
LGTM! ✅

danielmkm (approved):
Agree, think this can get merged

PR #97, opened 2023-09-21
Creating a Balancer API provider to fetch pool state for join txs;

lgahdl:
This Balancer API Provider is a utility module designed to provide data for b-sdk functions, initially it will provide pool state input for both join's and exit's, it's divided in "client" and the "modules",

Balancer API Client:
A Http Client that fetches the data from a specified url using a GraphQL query, the url is expected to be from a Beethoven X API or Balancer API, in this [link](https://github.com/beethovenxfi/beethovenx-backend/blob/v3-main/README.md#branching-and-deployment-environments) there are some valid urls.

Balancer API Modules:
As a first use case, the module "pool-state" was implemented to provide data for exit's and join's functions of the b-sdk, other modules will implement different data fetchings with different data structures that will be important for other b-sdk functionalities.


src/data/providers/balancer-api-provider/client/index.ts
```diff
@@ -0,0 +1,22 @@
+export class BalancerApiClient {
+  subgraphUrl: string;
+  chainId: number;
+  constructor(subgraphUrl: string, chainId: number) {
```
johngrantuk, line 4: `subgraphUrl` needs renamed.

johngrantuk:
Would be nice to add examples showing how this can be used, could add one for join and one for exit.

src/data/providers/balancer-api-provider/modules/join/index.ts
```diff
@@ -0,0 +1,88 @@
+import { BalancerApiClient } from '../../client';
+import { PoolState } from './types';
+
+export class JoinData {
```
johngrantuk, line 4: The same input data is used for Joins and Exits.

src/data/providers/balancer-api-provider/modules/join/types.ts
```diff
@@ -0,0 +1,14 @@
+import { Address } from "viem";
```
brunoguerios, line 1: You should be able to import `Address` from `/src/types` instead of viem. They are the same type, but it would be nice to keep the standard set on the rest of the code.

README.md
```diff
@@ -21,3 +21,63 @@ If your platform does not support one of the required features, it is also possi
 Testing requires access to an archive node for onchain quote comparisons. This can be done using Infura.
 
 `pnpm test`
+
+## Examples
+
+### Balancer Api Provider
+
+Joining with pool state:
```
brunoguerios, line 29: This example should be updated with the latest version of the code where JoinParser refactored.

README.md
```diff
@@ -21,3 +21,63 @@ If your platform does not support one of the required features, it is also possi
...
+
+    const balancerApi = new BalancerApi('https://api-v3.balancer.fi/', 1);
+    const poolState = await balancerApi.pools.fetchSimplePoolState('0x5f1d6874cb1e7156e79a7563d2b61c6cbce03150000200000000000000000586');
+    const joinParser = new JoinParser();
+    const poolJoin = joinParser.getJoin(poolState.type);
+    const queryResult = await weightedJoin.query(joinInput, poolState);
+
+```
+
+Exiting with pool state:
```
brunoguerios, line 48: The example below is describing a join flow, when it should be describing an exit, right?

src/entities/exit/parser.ts
brunoguerios, line 1: You should be able to remove this file, right?

src/entities/join/parser.ts
brunoguerios, line 1: Same here

brunoguerios (changes requested):

src/entities/exit/parser.ts
lgahdl: true, when I resolved conflicts with the join branch I accepted the merged branch change and it became an empty file, didn't notice that, thanks!

brunoguerios (commented):
LGTM 👍

src/data/providers/balancer-api/client/index.ts
```diff
@@ -0,0 +1,22 @@
+export class BalancerApiClient {
+  apiUrl: string;
+  chainId: number;
```
johngrantuk, line 3: Lets make this `ChainId` type from `src/utils/constants.ts`

src/data/providers/balancer-api/client/index.ts
```diff
@@ -0,0 +1,22 @@
...
+  constructor(apiUrl: string, chainId: number) {
+    this.apiUrl = apiUrl;
+    this.chainId = chainId;
+  }
+
+  async fetch(operationName: string, query: any, variables: any) {
```
johngrantuk, line 9: Anyway to tighten types of query and variables?

johngrantuk:
@lgahdl once the latest updates are pushed can you generate a changelog for the PR please.

package.json
```diff
@@ -39,6 +39,7 @@
         "pino-pretty": "^10.0.0",
         "rome": "12.1.3",
         "ts-node": "^10.9.1",
+        "tsconfig-paths": "^4.2.0",
```
johngrantuk, line 42: Whats this one for?

src/data/providers/balancer-api/client/index.ts
```diff
@@ -0,0 +1,31 @@
+import { ChainId } from '../../../../utils';
+
+export class BalancerApiClient {
+    apiUrl: string;
+    chainId: ChainId;
+    constructor(apiUrl: string, chainId: number) {
```
johngrantuk, line 6: Make `chainId: ChainId` instead?

src/data/providers/balancer-api/modules/pool-state/index.ts
```diff
@@ -0,0 +1,88 @@
...
+      }
+    }
+}`;
+
+  constructor(private readonly balancerApiClient: BalancerApiClient) {}
+
+  async fetchPoolState(id: string): Promise<PoolStateInput> {
+    const {
+      data: { poolGetPool },
+    } = await this.balancerApiClient.fetch('GetPool', this.poolStateQuery, {
```
johngrantuk, line 83: Is the `GetPool` (or even `operationName`) necessary? Can you explain what they are as I don't think I've seen it before.

src/data/providers/balancer-api/client/index.ts
```diff
@@ -0,0 +1,31 @@
...
+        this.apiUrl = apiUrl;
+        this.chainId = chainId;
+    }
+
+    async fetch(
```
johngrantuk, line 11: Is this meant to be flexible to allow different types of querys (apart from GetPool)? `variables: { id: string }` would probably limit that?

src/data/providers/balancer-api/modules/pool-state/index.ts
lgahdl: The "operationName" is not mandatory when you have only one function in the query, like we do, I can remove it, it's used to specify which function of the query will be executed

src/data/providers/balancer-api/client/index.ts
johngrantuk: How about changing so `BalancerApiClient.fetch` accepts the JSON query as the only param and `Pools.fetchPoolState` is responsible for putting that together?

lgahdl: This is a good idea, it will have only the requestQuery as a parameter on the fetch

package.json
lgahdl: I installed this to run the example files, but I removed the "example" script when I resolved conflicts in the package.json, I'll put it back

johngrantuk: Can it be removed now?

package.json
```diff
@@ -39,10 +42,14 @@
...
         "tsup": "^6.6.0",
         "typescript": "^5.0.4",
         "vite": "^4.4.2",
         "vitest": "^0.34.6"
     },
-    "packageManager": "^pnpm@8.6.0"
+    "packageManager": "^pnpm@8.6.0",
+    "resolutions": {
```
johngrantuk, line 52: Whats this for?

src/data/providers/balancer-api/client/index.ts
```diff
@@ -0,0 +1,24 @@
...
+export class BalancerApiClient {
+    apiUrl: string;
+    chainId: ChainId;
+    constructor(apiUrl: string, chainId: ChainId) {
+        this.apiUrl = apiUrl;
+        this.chainId = chainId;
+    }
+
+    async fetch(
+        requestQuery: {operationName?:string, query:string, variables?:any}
```
johngrantuk, line 12: Seems like your formatting/lint still isn't running correctly.

src/data/providers/balancer-api/modules/pool-state/index.ts
```diff
@@ -0,0 +1,90 @@
...
+            index
+          }
+        }
+      }
+    }
+}`;
+
+  constructor(private readonly balancerApiClient: BalancerApiClient) {}
+
+  async fetchPoolState(id: string): Promise<PoolStateInput> {
```
johngrantuk, line 80: Thinking ahead to next version, there won't be poolId. Is it possible to filter on address instead?

package.json
lgahdl: When running the examples, I was receiving this error:
```
Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/lgahdl/Documents/b-sdk/node_modules/.pnpm/get-port@6.1.2/node_modules/get-port/index.js from /Users/lgahdl/Documents/b-sdk/node_modules/.pnpm/@viem+anvil@0.0.6/node_modules/@viem/anvil/dist/cjs/pool/createPool.js not supported.
```
This error occurs because "@viem/anvil" imports other package "get-port" using cjs require, and get-port v6+ allows only ES import, if we had webpack or babel, this would be very easier to fix, but in my case I could only fix making the "get-port"version 5.1.1, I don't think this is the best solution, so I asked for Alberto's help, he may have a better approach for that.

src/data/providers/balancer-api/modules/pool-state/index.ts
lgahdl: With the API we have now, it's not possible with the query "poolGetPools", but is possible with the query "poolGetPools", the problem is that "poolGetPools" returns a pool model that cannot retrieve the token index, so we have 2 options:

1 - Create a query similar to poolGetPool that accept address
2 - Adapt the query poolGetPools to retrieve index

But either way we will have to make changes in the API.

johngrantuk: Thanks for this info. Will create an Asana ticket to handle this.

johngrantuk (approved):

PR #98, opened 2023-09-26
fix: Circular deps.

johngrantuk:
Error on current [join/exit PR](https://github.com/balancer/b-sdk/pull/91) seem to be related to circular dependency loop issues. I used https://www.npmjs.com/package/dpdm package to analyze where these could be occurring and fixed most by importing directly instead of from module.


brunoguerios (approved):
I thought it was something related to imports. Thanks for sharing the tool link. Will save it over here :)

PR #99, opened 2023-09-27
Add Add/Remove Liquidity Nested

brunoguerios:
Done:
- AddLiquidityNested
- AddLiquidityNested with native asset
- RemoveLiquidityNestedProportional
- RemoveLiquidityNestedProportional with native asset
- RemoveLiquidityNestedSingleToken
- RemoveLiquidityNestedSingleToken with native asset

Pending:
- Examples
- Update readme

Notes:
- relies on locally deployed relayer, so github checks will fail until we add these steps to the flow or SC deploys the new relayer version to mainnet


src/entities/encoders/stable.ts
johngrantuk, line 1: Just to confirm - this is for ComposableStable rather than other/older Stable pool types?

src/utils/constants.ts
```diff
@@ -124,7 +124,11 @@ export const BALANCER_POOL_DATA_QUERIES_ADDRESSES: Record<number, Address> = {
     [ChainId.ZKEVM]: '0xF24917fB88261a37Cc57F686eBC831a5c0B9fD39',
 };
 
-export const BALANCER_VAULT = '0xBA12222222228d8Ba445958a75a0704d566BF2C8';
+export const BALANCER_VAULT: Address =
+    '0xBA12222222228d8Ba445958a75a0704d566BF2C8';
+
+export const BALANCER_RELAYER =
```
johngrantuk, line 130: This will likely have to be per network.

src/entities/nestedJoin/index.ts
```diff
@@ -0,0 +1,270 @@
...
+import { Slippage } from '../slippage';
+import { Token } from '../token';
+import { BALANCER_RELAYER, CHAINS, getPoolAddress } from '../../utils';
+import { Relayer } from '../relayer';
+import { parseNestedJoinArgs } from '../utils/parseNestedJoinArgs';
+import { TokenAmount } from '../tokenAmount';
+import { balancerRelayerAbi, bathcRelayerLibraryAbi } from '../../abi';
+
+export type NestedJoinInput = {
+    amountsIn: {
```
johngrantuk, line 18: This format of user input is based off the discussion we had previously with Daniel?

src/entities/nestedJoin/index.ts
```diff
@@ -0,0 +1,270 @@
...
+    }[];
+    chainId: number;
+    rpcUrl: string;
+    testAddress: Address;
+    useNativeAssetAsWrappedAmountIn?: boolean;
+    fromInternalBalance?: boolean;
+};
+
+export type JoinStep = {
+    action: 'join'; // TODO: are we supporting other actions, such as swaps?
```
johngrantuk, line 30: Without Linear pool support then join only should be fine I think.

src/entities/nestedJoin/index.ts
```diff
@@ -0,0 +1,270 @@
...
+    slippage: Slippage;
+    sender: Address;
+    recipient: Address;
+    relayerApprovalSignature?: Hex;
+};
+
+const outputRefOffset = 100n;
+
+export class NestedJoin {
+    async query(
```
johngrantuk, line 85: I think in general this function could be refactored to smaller functions that will make it clearer, e.g.  `getPoolCall`, `getMaxAmountsIn()`, etc

src/entities/nestedJoin/index.ts
```diff
@@ -0,0 +1,270 @@
...
+            chain: CHAINS[input.chainId],
+        });
+
+        // sort pools by ascending level
+        // then go from bottom pool to up filling out input and output amounts from joinSteps
+        // input at level 0 are either amountsIn provided or 0n
+        // input at following levels can be amountsIn provided, output of the previous level or 0n
+        // output at max level is the bptOut
+
+        const stepsSortedByLevel = nestedPoolState.joinSteps.sort(
```
johngrantuk, line 100: Feels like `NestedPoolState` data structure could be changed to be more useful as we're really just manipulating it before its being used. Might be better to have each pool contain its joinSteps and level? something like:
```
    pools: {
        id: Hex;
        address: Address;
        type: string;
        level: number;
        joinSteps: JoinStep[]; // each token should have at least one
    }[];
  
```

src/entities/encoders/stable.ts
```diff
@@ -0,0 +1,123 @@
...
+    TOKEN_IN_FOR_EXACT_BPT_OUT = 2,
+}
+
+export enum StablePoolExitKind {
+    EXACT_BPT_IN_FOR_ONE_TOKEN_OUT = 0,
+    EXACT_BPT_IN_FOR_TOKENS_OUT = 1,
+    BPT_IN_FOR_EXACT_TOKENS_OUT = 2,
+}
+
+// TODO: rename functions after deciding on the naming convention
```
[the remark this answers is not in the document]
brunoguerios: No, that's likely a left over from a previous version of the code that I copied over.
Thanks for pointing it out. I'll remove it over here 👍

src/entities/encoders/stable.ts
brunoguerios: This is actually the code for Stable encoder, which I realized only after a while.
I left it as is because coincidentally both behave the same for the types of joins I'm using.
In any case, I already have a TODO on the code to add a ComposableStable encoder and replace the usage within nested pools. 👍

src/utils/constants.ts
brunoguerios: You're right - I'll go ahead and update it to be per network 👍

src/entities/nestedJoin/index.ts
brunoguerios: Yes, we agreed on not using `TokenAmount` externally, right? I'm just not sure if we're going to need `decimals` or any other token info here. I think more data requirements will pop up once I start adding complexity, such as duplicate tokens that will require amount manipulation before building the nested calls.

src/entities/nestedJoin/index.ts
brunoguerios: Totally agreed. I thought about leaving this kind of refactor to after adding more test scenarios, because I think the code will change a bit until then.

src/entities/nestedJoin/index.ts
brunoguerios: Yeap, I agree it would be nice to move data to inside the pools. 👍

brunoguerios: And thinking a bit more about the data structure, since `level` was pulled out to the pool and `isTop` is not actually being used at the moment, I'd say we could move to something like this:
```
    pools: {
        id: Hex;
        address: Address;
        type: string;
        level: number;
        nestedTokens: {
            address: Address;
            balance: bigint;
            decimals: number;
            index: number;
            action: 'join';
        }[];
    }[];
```

johngrantuk: If you rename `nestedTokens` to just `tokens` it keeps consistency and ends up being basically the same as the normal join but with a bit of extra data.

brunoguerios: Ok 👍  We could even remove the action 'join' while we don't have any other actions. And if/when we add them in the future, they could be a pool rather than a token data.

src/entities/encoders/stable.ts
johngrantuk, line 1: As far as I can see this isn't used anywhere? And we don't planning on supporting earlier Stables so lets delete this.

src/utils/constants.ts
```diff
@@ -124,7 +124,12 @@ export const BALANCER_POOL_DATA_QUERIES_ADDRESSES: Record<number, Address> = {
     [ChainId.ZKEVM]: '0xF24917fB88261a37Cc57F686eBC831a5c0B9fD39',
 };
 
-export const BALANCER_VAULT = '0xBA12222222228d8Ba445958a75a0704d566BF2C8';
+export const BALANCER_VAULT: Address =
+    '0xBA12222222228d8Ba445958a75a0704d566BF2C8';
+
+export const BALANCER_RELAYER: Record<number, Address> = {
+    [ChainId.MAINNET]: '0xa8d297D643a11cE83b432e87eEBce6bee0fd2bAb',
```
johngrantuk, line 131: Just a reminder before we merge that this needs updated for other networks.

johngrantuk:
@brunoguerios - as this needs a new Relayer I think it might be best to mark as Draft until its deployed then release after that, WDYT?

test/nestedJoin.integration.test.ts
```diff
@@ -0,0 +1,429 @@
...
+    nestedJoin,
+    nestedPoolFromApi,
+    client,
+    useNativeAssetAsWrappedAmountIn = false,
+}: TxInput) => {
+    const joinInput: NestedJoinInput = {
+        amountsIn,
+        chainId,
+        rpcUrl,
+        testAddress,
```
johngrantuk, line 250: This is throwing a type error for me.

src/entities/nestedExit/index.ts
```diff
@@ -0,0 +1,127 @@
...
+
+        const peekedValues = await doQueryNestedExit(
+            input.chainId,
+            input.rpcUrl,
+            input.accountAddress,
+            encodedMulticall,
+            tokensOut.length,
+        );
+
+        console.log('peekedValues ', peekedValues);
```
brunoguerios, line 57: A reminder so I don't forget to remove this debug log

src/entities/nestedJoin/getNestedJoinCalls.ts
```diff
@@ -0,0 +1,70 @@
+import { Token } from '../token';
+import { getPoolAddress } from '../../utils';
+import { NestedJoinInput, NestedJoinCall } from './types';
+import { NestedPoolState } from '../types';
+
+export const getNestedJoinCalls = (
```
johngrantuk, line 6: What do you think about changing the naming for this? I found it a bit confusing because its not really creating any calls. Really its getting pool data that can be used to build a call later. Specifically a query call with min out of 0.

src/entities/nestedJoin/getNestedJoinCalls.ts
```diff
@@ -0,0 +1,70 @@
...
+            chainId: chainId,
+            useNativeAssetAsWrappedAmountIn:
+                useNativeAssetAsWrappedAmountIn ?? false,
+            sortedTokens,
+            poolId: pool.id,
+            poolType: pool.type,
+            kind: 0,
+            sender: accountAddress,
+            recipient: accountAddress,
+            maxAmountsIn: sortedTokens.map((token) => {
```
johngrantuk, line 40: Can you break this logic into a separate function that is commented a bit to explain. Would make it easier to follow.
```
Inputs will be:
  - bottom pool:
    - amountsIn if provided by user
    - 0 if no user amount
  - middle pool/top pool:
    - output of the previous level
```

src/entities/nestedJoin/getNestedJoinCalls.ts
```diff
@@ -0,0 +1,70 @@
...
+                        isRef: true,
+                    };
+                } else {
+                    return {
+                        amount: 0n,
+                        isRef: false,
+                    };
+                }
+            }),
+            minBptOut: 0n,
```
johngrantuk, line 64: I think its worth highlighting this a bit in a comment and also in function comment. Just to make it clear that this is for query only.

src/entities/nestedJoin/getNestedJoinCalls.ts
```diff
@@ -0,0 +1,70 @@
...
+            .map((t) => new Token(chainId, t.address, t.decimals));
+        calls.push({
+            chainId: chainId,
+            useNativeAssetAsWrappedAmountIn:
+                useNativeAssetAsWrappedAmountIn ?? false,
+            sortedTokens,
+            poolId: pool.id,
+            poolType: pool.type,
+            kind: 0,
+            sender: accountAddress,
```
johngrantuk, line 38: I think the multicall could be more efficient by using relayer and internal balances right? (Can be done in a future PR but just wanted to clarify at this stage)

src/entities/nestedJoin/parseNestedJoinCall.ts
```diff
@@ -0,0 +1,82 @@
...
+    recipient,
+    maxAmountsIn,
+    minBptOut,
+    fromInternalBalance,
+    outputReferenceKey,
+}: NestedJoinCall) => {
+    // replace wrapped token with native asset if needed
+    let tokensIn = [...sortedTokens];
+    let value = 0n;
+    if (chainId && useNativeAssetAsWrappedAmountIn) {
```
johngrantuk, line 26: I don't think we need to check chainId here?

src/entities/nestedJoin/parseNestedJoinCall.ts
```diff
@@ -0,0 +1,82 @@
...
+        tokensIn = replaceWrapped([...sortedTokens], chainId);
+        const nativeAssetIndex = tokensIn.findIndex((t) =>
+            t.isSameAddress(ZERO_ADDRESS),
+        );
+        if (nativeAssetIndex > -1) {
+            value = maxAmountsIn[nativeAssetIndex].amount;
+        }
+    }
+
+    const poolAddress = getPoolAddress(poolId);
```
johngrantuk, line 36: Since we're already getting poolAddress in `getNestedJoinCalls` should we use that as another param?

src/entities/nestedJoin/types.ts
```diff
@@ -0,0 +1,48 @@
...
+    outputReferenceKey: bigint;
+};
+
+export type NestedJoinQueryResult = {
+    calls: NestedJoinCall[];
+    bptOut: TokenAmount;
+};
+
+export type NestedJoinCallInput = NestedJoinQueryResult & {
+    chainId: number;
```
johngrantuk, line 43: We also have access to chainId in the `calls` so could reuse that?

test/nestedJoin.integration.test.ts
```diff
@@ -0,0 +1,432 @@
...
+
+        testAddress = (await client.getAddresses())[0];
+
+        poolAddress = '0x08775ccb6674d6bdceb0797c364c2653ed84f384'; // WETH-3POOL-BPT
+
+        // get pool state from api
+        nestedPoolFromApi = await api.getNestedPool(poolAddress);
+
+        // setup join helper
+        nestedJoin = new NestedJoin();
```
johngrantuk, line 107: I normally find tests are a good place to see how functionality can be used. Would be good to keep all the test setup, etc together (or even better in a separate reusable helper) and the core functionality before/after. Similar for doTransaction. Does that make sense?

test/nestedJoin.integration.test.ts
```diff
@@ -0,0 +1,432 @@
...
+        })
+            .extend(publicActions)
+            .extend(walletActions);
+
+        testAddress = (await client.getAddresses())[0];
+
+        poolAddress = '0x08775ccb6674d6bdceb0797c364c2653ed84f384'; // WETH-3POOL-BPT
+
+        // get pool state from api
+        nestedPoolFromApi = await api.getNestedPool(poolAddress);
```
johngrantuk, line 104: One thing in terms of UX - how does a user know if they need a nested join? 
We maybe need some helper that can show ifNested and what tokens can be used to join.

test/nestedJoin.integration.test.ts
```diff
@@ -0,0 +1,432 @@
...
+                ...tokensIn.map((t) => t.address),
+                queryResult.bptOut.token.address,
+            ],
+            client,
+            testAddress,
+            to,
+            call,
+            value,
+        );
+
```
johngrantuk, line 296: Should add a test for `value`. Mainly to make sure there is no value set when it shouldn't be as this would be bad.

src/entities/nestedJoin/getNestedJoinCalls.ts
```diff
@@ -0,0 +1,70 @@
...
+                } else {
+                    return {
+                        amount: 0n,
+                        isRef: false,
+                    };
+                }
+            }),
+            minBptOut: 0n,
+            fromInternalBalance: fromInternalBalance ?? false,
+            outputReferenceKey: BigInt(poolsSortedByLevel.indexOf(pool)) + 100n,
```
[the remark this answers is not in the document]
brunoguerios: It's kind of a left over from the previous implementation where I was avoiding to use ref equals zero. I don't actually remember why. I'll remove it over here.

src/entities/nestedJoin/getNestedJoinCalls.ts
brunoguerios: We should be able to use the relayer to reduce gas cost 👍 
But chaining actions with internal balances require each action to have both `from` and `to` internal balance available, which is not the case for `joins` and `exits`. The previous implementation used it mainly with `swaps`.
The best we can do is to have the first join action with `fromInternalBalance = true` and the last exit action with `toInternalBalance = true` - I'll go ahead and make that change.

src/entities/nestedJoin/getNestedJoinCalls.ts
brunoguerios: The way I was interpreting the result was as non-encoded calls, which would then be formatted (through getParsedCalls) and then encoded, but I agree that it's a bit confusing.
And I agree it should mention something about being query data, since it's setting minAmountsOut to 0.
With that in mind, I'll rename it to `getQueryCallData` (no need to keep NestedJoin within the name, since it's being used internally only)
Let me know if you have a better suggestion 😉

src/entities/nestedJoin/index.ts
```diff
@@ -0,0 +1,123 @@
...
+export class NestedJoin {
+    async query(
+        input: NestedJoinInput,
+        nestedPoolState: NestedPoolState,
+    ): Promise<NestedJoinQueryResult> {
+        const calls = getNestedJoinCalls(input, nestedPoolState);
+
+        const parsedCalls = calls.map((call) => parseNestedJoinCall(call));
+
+        const encodedCalls = parsedCalls.map((parsedCall) =>
```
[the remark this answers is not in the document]
brunoguerios: I think so 👍 
I was trying to break into very granular steps, but there's indeed not benefit on exposing the formatted output we get from `parseNestedJoinCall`.
I'll move the encoding inside and rename it to `encodeCallData`.

src/entities/nestedJoin/index.ts
```diff
@@ -0,0 +1,123 @@
...
+                functionName: 'joinPool',
+                args: parsedCall.args,
+            }),
+        );
+
+        // append peek call to get bptOut
+        const peekCall = Relayer.encodePeekChainedReferenceValue(
+            Relayer.toChainedReference(
+                calls[calls.length - 1].outputReferenceKey,
+                false,
```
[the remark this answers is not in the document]
brunoguerios: Good question. I'm using non-temp refs based on the previous peek implementation, but I'm not sure it's needed. I'll check and let you know what I find out.

src/entities/nestedJoin/types.ts
```diff
@@ -0,0 +1,48 @@
+import { Address, Hex } from '../../types';
+import { Slippage } from '../slippage';
+import { Token } from '../token';
+import { TokenAmount } from '../tokenAmount';
+
+export type NestedJoinInput = {
+    amountsIn: {
```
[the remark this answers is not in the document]
brunoguerios: The main issue with `TokenAmount` is that it has attributes that could cause some confusion, such as `wrapped`.
The minimal info we need from the user is token `address` and `rawAmount`, because we can get `decimals` from the token provided by the api (i.e. poolState), but if I remember correctly @danielmkm wanted to keep decimals there to keep consistency.
In the end that was the format that made most sense.
Let me know if you think there's a better option.

test/nestedJoin.integration.test.ts
brunoguerios: I agree. I'll refactor it over here.

test/nestedJoin.integration.test.ts
brunoguerios: That's an interesting question.
I think that a method to check if a pool is nested or not makes more sense at the api level.
Same as to find out which tokens could be used to join.
We could add a helper within the SDK that requests that kind info from the API and forwards that to the user.
Is that what you have in mind?

johngrantuk: Yeah, might be a good time to take advantage of the API. 
Maybe the user shouldn't have to think about nested vs normal join at all.
Might be possible to have a single `getPool` endpoint that will return a list of tokens that can be used to join along with the data needed to construct the call. Then the joinPool can handle which method to join with. One issue would be dealing with constraints around joining with different tokens.
```
{
  id,
  address,
  type,
  joinTokens: [] --- list of tokens that can be used to join, including nested
  poolTokens
  nested: {
    pools: [{id, type, etc}, {...}] --- data needed to construct join (same as now)
  }
}
```

src/entities/nestedExit/getNestedExitCalls.ts
```diff
@@ -0,0 +1,193 @@
...
+        bptAmountIn,
+        chainId,
+        accountAddress,
+        tokenOut,
+        useNativeAssetAsWrappedAmountOut = false,
+        toInternalBalance = false,
+    }: NestedExitInput,
+    { pools }: NestedPoolState,
+): { bptAmountIn: TokenAmount; calls: NestedExitCall[] } => {
+    const isProportional = tokenOut === undefined;
```
johngrantuk, line 18: This is common for `getNestedExitCalls` and `parseNestedExitCall`, we could do it early and explicitly in NestedExit.query and pass a boolean or enum as a param to those functions which would make it clearer.

src/entities/nestedExit/getNestedExitCalls.ts
```diff
@@ -0,0 +1,193 @@
...
+        /**
+         * Overall logic to build sequence of single token exit calls:
+         * 1. Go BOTTOM-UP building exit path to tokenOut
+         * 2. Go through exit path filling out input amounts and output refs
+         * 3. Inputs will be bptAmountIn provided or output of the previous level
+         * 4. Output at bottom level is the amountOut
+         */
+
+        // sort pools by descending level
+        poolsTopDown = pools.sort((a, b) => b.level - a.level);
```
johngrantuk, line 51: How about bringing all this logic together into `getSingleTokenExitCalls` which keeps it more naturally grouped? Same with above.

src/entities/nestedExit/index.ts
```diff
@@ -0,0 +1,127 @@
...
+} from './types';
+import { NestedPoolState } from '../types';
+import { doQueryNestedExit } from './doQueryNestedExit';
+import { getNestedExitCalls } from './getNestedExitCalls';
+import { parseNestedExitCall } from './parseNestedExitCall';
+import { getPeekCalls } from './getPeekCalls';
+
+export class NestedExit {
+    async query(
+        input: NestedExitInput,
```
johngrantuk, line 20: Could we make the `NestedExitInput` a Union of the two exit types, Proportional and SingleToken.

src/entities/nestedExit/getNestedExitCalls.ts
```diff
@@ -0,0 +1,193 @@
...
+        const topPool = poolsTopDown[0];
+
+        // Go BOTTOM-UP building exit path to tokenOut
+        const exitPath: NestedPool[] = [];
+        let tokenOutByLevel = tokenOut;
+        while (tokenOutByLevel !== topPool.address) {
```
johngrantuk, line 57: Could probably be split into a separate a separate function with a description of whats happening.

src/entities/nestedExit/getNestedExitCalls.ts
```diff
@@ -0,0 +1,193 @@
...
+            const currentPool = poolsTopDown.find(
```
johngrantuk, line 58: Should handle/throw when pool isn't found just to be clear.

johngrantuk, line 58: Just so we cover it - this is always finding the first pool. So in the case there are more than 1 pools with the token it should still work but it'll always find the same path? Can you see any issues here?

src/entities/nestedExit/getNestedExitCalls.ts
```diff
@@ -0,0 +1,193 @@
...
+                (p) =>
+                    p.address !== tokenOutByLevel && // prevents pools with BPT as token to be picked up incorrectly
```
johngrantuk, line 60: Could you give me an example case of this when you get a chance?

test/nestedJoin.integration.test.ts
brunoguerios: Yeah, that feels more like a business than a technical decision to me.
We seem to be leaning towards avoiding one-size-fits-all approaches elsewhere (e.g. vault, etc..), so I'd assume we would follow the same idea here.
I like a pool model that returns everything you need to perform either types of joins, but separated in a way that it's clear what's it for.
With that in mind, I'd expect something like this:
```
{
  id,
  address,
  type,
  poolTokens --- for normal join
  nested: {
    pools: [{id, type, etc}, {...}] --- data needed to construct join (same as now)
    joinTokens: [] --- list of tokens that can be used to join (nested only)
  }
}
```
Where `nested` is optional and exists only for nested pools.

src/entities/nestedExit/getPeekCalls.ts
```diff
@@ -0,0 +1,49 @@
...
+        peekCalls.push(
+            Relayer.encodePeekChainedReferenceValue(
+                Relayer.toChainedReference(
+                    lastCall.outputReferenceKeys[0],
+                    false,
+                ),
+            ),
+        );
+    } else {
+        calls.forEach((call) => {
```
johngrantuk, line 27: Some comments here would be useful. Finding it a bit hard to reason.

test/nestedJoin.integration.test.ts
brunoguerios: Maybe one thing that would help us answer how this should look like is: how will the FE present different options of joins on the UI?
- a single page with all join tokens (normal + nested) -> then the first approach makes more sense.
- two separate pages/tabs with each type of join and it's inputs -> then the second approach makes more sense.

johngrantuk: Thats a good question. I'll make a note to bring this up with Daniel in the standup on Monday.

src/entities/nestedExit/getNestedExitCalls.ts
brunoguerios: The code will still work and I'd say that it should find the same path if the api response returns pools in the same order (which I'm not sure is enforced).
But taking a step back, do we want to allow this kind of behavior?
I'd say that the api should not provide nested join/exit data for pools that are outside of the scope.
In other words, I'd say it's the api responsibility to prevent cases like that from ever reaching the SDK.
What do you think?

src/entities/nestedExit/getNestedExitCalls.ts
brunoguerios: Sure! The pool being used on the tests (WETH/3-POOL) has a case where this happens.
When exiting to DAI for example, the first iteration will pick 3-POOL as the "bottom" pool and update tokenOutByLevel to 3-POOL-BPT. Since 3-POOL-BPT is contained on both WETH/3-POOL and 3-POOL itself, simply checking if the pool contains that token could result in the bottom pool being picked up again.
That's why I have to filter out pools that have tokenOutByLevel as it's own address.
Let me know if that's still not clear 😅

src/entities/nestedJoin/getNestedJoinCalls.ts
brunoguerios: Actually `callData` didn't seem to represent an array of calls attributes, so I decided to go with `getQueryCallsAttributes`.

src/entities/nestedJoin/getNestedJoinCalls.ts
brunoguerios: That's not actually possible, because this is not entirely determined by the level of the pool (bottom vs top).
Consider the WETH/3-POOL case. The top pool has a WETH amountIn provided by the user.
I'll try to add some comments and update some variables, in order to make it easier to follow.
Let me know if that's still not enough and we'll think of something else.

src/entities/nestedJoin/index.ts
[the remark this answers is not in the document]
brunoguerios: Non-temp refs work as expected 👍 
Regarding reusing the existing ref, we could find a way to merge `encodeCalls` with `getPeekCalls`, but I believe it's better to keep both logics separated, since encodeCalls is reused on `query` and `buildCall`.

test/nestedJoin.integration.test.ts
brunoguerios: Registering over here that we decided to keep both regular and nested joins on separate flows, which means that the second approach, with `joinTokens` within `nested` object makes more sense.
Since this format will be used for `exits` as well, should we rename `joinTokens` by `mainTokens`?

johngrantuk: Thanks Bruno. Yeah, lets keep consitency.

src/entities/nestedJoin/getNestedJoinCalls.ts
brunoguerios: Regarding `from/to internal balance`, on the previous implementation, we only used for intermediary calls. A join with fromInternalBalance where the user's balance was originally internal was not supported.
Is this something we should support on b-sdk?
Being more concrete:
- example of flow we supported: non-internal -> internal -> internal -> non-internal
- example of flow we didn't support: internal -> internal -> non-internal
I ask because it adds a bit of complexity on tests. Nothing too complex, but worth checking before implementing.

src/entities/nestedExit/getNestedExitCalls.ts
brunoguerios: I actually need this outside the if/else because it's being used to build `bptAmountIn` TokenAmount.

test/nestedJoin.integration.test.ts
```diff
@@ -0,0 +1,528 @@
...
+            functionName: 'grantRole',
+            args: [role, BALANCER_RELAYER[chainId]],
+        });
+    }
+    await client.stopImpersonatingAccount({
+        address: balancerDaoAddress,
+    });
+};
+
+export const approveRelayer = async (
```
johngrantuk, line 513: Resued across the tests so could be moved to a shared helper.

test/nestedExit.integration.test.ts
```diff
@@ -0,0 +1,476 @@
...
+
+function assertResults(
+    transactionReceipt: TransactionReceipt,
+    bptAmountIn: TokenAmount,
+    amountsOut: TokenAmount[],
+    balanceDeltas: bigint[],
+    slippage: Slippage,
+    minAmountsOut: TokenAmount[],
+) {
+    expect(transactionReceipt.status).to.eq('success');
```
johngrantuk, line 462: I think we can change this/doTransaction so we test both value and tokenAmounts for all the exit type. E.g. when its native exit we should check that value has updated correctly and that the wrappedToken is 0. Let me know if you think thats worthwhile or not?

test/lib/utils/helper.ts
```diff
@@ -24,6 +24,18 @@ export const approveToken = async (
     token: Address,
     amount = MAX_UINT256, // approve max by default
 ): Promise<boolean> => {
+    const allowance = await client.readContract({
```
johngrantuk, line 27: Break this out into a separate `hasAllowance` type helper.

src/entities/nestedJoin/types.ts
```diff
@@ -0,0 +1,48 @@
...
+        address: Address;
+        rawAmount: bigint;
+    }[];
+    chainId: number;
```
johngrantuk, line 11: Can make this `ChainId` type.

src/entities/nestedJoin/types.ts
```diff
@@ -0,0 +1,48 @@
...
+    rpcUrl: string;
+    accountAddress: Address;
+    useNativeAssetAsWrappedAmountIn?: boolean;
+    fromInternalBalance?: boolean;
+};
+
+export type NestedJoinCallAttributes = {
+    chainId: number;
```
johngrantuk, line 19: Same as above.

src/entities/nestedJoin/types.ts
```diff
@@ -0,0 +1,48 @@
...
+    useNativeAssetAsWrappedAmountIn: boolean;
+    sortedTokens: Token[];
+    poolId: Hex;
+    poolAddress: Address;
+    poolType: string;
+    kind: number;
```
johngrantuk, line 25: Could maybe make an enum type here to make it easier to follow what the different kinds are?

src/entities/nestedExit/types.ts
```diff
@@ -0,0 +1,51 @@
+import { Address, Hex } from '../../types';
+import { Slippage } from '../slippage';
+import { Token } from '../token';
+import { TokenAmount } from '../tokenAmount';
+
+export type NestedProportionalExitInput = {
+    bptAmountIn: bigint;
+    chainId: number;
```
johngrantuk, line 8: Could make this `ChainId` type.

src/entities/nestedExit/types.ts
```diff
@@ -0,0 +1,51 @@
...
+    useNativeAssetAsWrappedAmountOut?: boolean;
+    toInternalBalance?: boolean;
+};
+
+export type NestedSingleTokenExitInput = NestedProportionalExitInput & {
+    tokenOut: Address;
+};
+
+export type NestedExitCallAttributes = {
+    chainId: number;
```
johngrantuk, line 20: Same as above.

src/entities/nestedExit/index.ts
```diff
@@ -0,0 +1,147 @@
...
+
+        return {
+            call,
+            to: BALANCER_RELAYER[input.callsAttributes[0].chainId],
+            minAmountsOut,
+        };
+    }
+}
+
+const validateInputs = (
```
johngrantuk, line 122: What do you think about refactoring this a bit to be super clear? Could have two further functions: `validateProportional` and `validateSingleToken`.

src/entities/nestedExit/encodeCalls.ts
```diff
@@ -0,0 +1,119 @@
...
+                case 'ComposableStable':
+                    userData =
+                        ComposableStableEncoder.exitProportional(_bptAmountIn);
+                    break;
+                default:
+                    throw new Error('Unsupported pool type');
+            }
+
+            outputReferences = outputReferenceKeys.map((k) => {
+                const tokenIndex = k % 10n;
```
johngrantuk, line 59: Could you explain what this is for?

src/entities/nestedExit/encodeCalls.ts
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        const _bptAmountIn = bptAmountIn.isRef
+            ? Relayer.toChainedReference(bptAmountIn.amount)
+            : bptAmountIn.amount;
+
+        let userData: Hex;
+        let outputReferences: { index: bigint; key: bigint }[] = [];
+
+        if (isProportional) {
```
johngrantuk, line 45: I do think it could be useful to further modularise this code by having separate functions for proportional/singleToken specific operations. Would make it easier to read because we know what those functions should be doing. WDYT?

src/entities/nestedExit/getQueryCallsAttributes.ts
```diff
@@ -0,0 +1,201 @@
...
+    NestedSingleTokenExitInput,
+    NestedExitCallAttributes,
+} from './types';
+import { NestedPool, NestedPoolState } from '../types';
+import { TokenAmount } from '../tokenAmount';
+import { Address } from '../../types';
+
+export const getQueryCallsAttributes = (
+    input: NestedProportionalExitInput | NestedSingleTokenExitInput,
+    { pools }: NestedPoolState,
```
johngrantuk, line 13: Can't we just change this to be: `pools: NestedPool[]`?

src/entities/nestedExit/getQueryCallsAttributes.ts
```diff
@@ -0,0 +1,201 @@
...
+        useNativeAssetAsWrappedAmountOut = false,
+        toInternalBalance = false,
+    } = input;
+    let callsAttributes: NestedExitCallAttributes[];
+
+    // sort pools by descending level
+    const poolsTopDown = pools.sort((a, b) => b.level - a.level);
+
+    if (isProportional) {
+        /**
```
johngrantuk, line 32: These comments are probably better as part of the actual functions now.

src/entities/nestedExit/getQueryCallsAttributes.ts
```diff
@@ -0,0 +1,201 @@
...
+    const calls: NestedExitCallAttributes[] = [];
+    for (const pool of poolsSortedByLevel) {
+        const sortedTokens = pool.tokens
+            .sort((a, b) => a.index - b.index)
+            .map((t) => new Token(chainId, t.address, t.decimals));
+
+        const sortedTokensWithoutBpt = sortedTokens.filter(
+            (t) => !t.isSameAddress(pool.address),
+        );
+        const upperLevelCall = calls.find((call) =>
```
johngrantuk, line 91: Is `upperLevel` right here?

src/entities/nestedJoin/getQueryCallsAttributes.ts
```diff
@@ -0,0 +1,83 @@
...
+            useNativeAssetAsWrappedAmountIn:
+                useNativeAssetAsWrappedAmountIn ?? false,
+            sortedTokens,
+            poolId: pool.id,
+            poolAddress: pool.address,
+            poolType: pool.type,
+            kind: 0,
+            sender: accountAddress,
+            recipient: accountAddress,
+            maxAmountsIn: sortedTokens.map((token) => {
```
johngrantuk, line 41: I still think this could benefit from being split into a separate function, e.g. `getMaxAmountsIn`.

src/entities/nestedJoin/getNestedJoinCalls.ts
johngrantuk: We definitely don't need to support that new user case.

test/nestedJoin.integration.test.ts
brunoguerios: Will we wait for the new relayer to be deployed before merging this PR? I didn't pay much attention to organizing these helper methods because I was expecting they would eventually be removed.
I'll refactor just in case 👍

test/nestedJoin.integration.test.ts
```diff
@@ -0,0 +1,528 @@
...
+                {
+                    address: '0xdac17f958d2ee523a2206206994597c13d831ec7', // USDT
+                    decimals: 6,
+                },
+            ],
+        };
+    }
+}
+
+export const grantRoles = async (
```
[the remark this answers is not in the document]
brunoguerios: Same as above 👍

test/lib/utils/helper.ts
brunoguerios: I was annoyed by the fact that I had to comment out token approvals for USDT because it fails if you try to approve it twice, so I added that, but I agree it makes sense to do it on a separate function.
I'll refactor over here 👍

src/entities/nestedJoin/types.ts
brunoguerios: Sure! It would be interesting to create a separate task to refactor this throughout b-sdk because it's being used as number everywhere.

src/entities/nestedExit/encodeCalls.ts
brunoguerios: I set outputReferenceKey in a way that its last digit represented the index of the token in the pool, so I was reusing that here to get the index out, but I agree it's non-ideal to have this kind of hidden logic in the code. I'll go ahead and update outputReferences so this info is more explicit and doesn't cause confusion.

src/entities/nestedExit/encodeCalls.ts
brunoguerios: Agree 👍

src/entities/nestedExit/getQueryCallsAttributes.ts
brunoguerios: Yeap 👍

test/nestedJoin.integration.test.ts
johngrantuk: I think we should wait otherwise the functionality won't work.

test/nestedJoin.integration.test.ts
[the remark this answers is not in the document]
johngrantuk: No need to refactor if we will remove it after relayer is deployed. 
Might be worth keeping this snippet in Notion though as it could be useful in future.

src/entities/nestedExit/getQueryCallsAttributes.ts
brunoguerios: It's actually trying to find a previous call that has its output as the current call input.
We could have a more descriptive name, such as `previousCall` or `previousNestedCall`
Or if that's too complex to describe on a variable name, we could add more details on a comment.

brunoguerios: I renamed this upperLevelCall to `previousCall` and refactored it into `getBptAmountIn` logic to make it easier to understand. Let me know what you think.

test/nestedJoin.integration.test.ts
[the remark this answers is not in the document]
brunoguerios: I moved them to a relayerHelper file. I fear that they will be lost within Notion. In the source code, they are easier to find.
If our package is tree-shakeable and these methods are never used in production, they should not affect the b-sdk size anyway.
Let me know if you still think they should be deleted once they are no longer needed for the current tests.

test/nestedExit.integration.test.ts
brunoguerios: For exits we don't ever add `value`, so we don't need to check it, right?
For nestedJoin, it's already checking `value` for all tests and it's checking balance deltas for amountsIn/Out (for ETH it also discounts gas cost).
Given all that matches properly, I don't see a benefit on checking WETH balance deltas as well.
Let me know if you think we could be missing some weird edge cases with that assumption.

package.json
```diff
@@ -25,6 +23,7 @@
         "test:ci": "vitest run",
         "changeset": "changeset",
         "changeset:release": "pnpm build && changeset publish",
+        "anvil": "anvil --fork-url $(. ./.env && echo $ETHEREUM_RPC_URL)",
```
johngrantuk, line 26: This is only needed until Relayer V6 is deployed right? Note to remove once we have updated tests.

src/entities/nestedJoin/encodeCalls.ts
```diff
@@ -0,0 +1,96 @@
+import { Hex } from '../../types';
+import { ZERO_ADDRESS } from '../../utils';
+import { WeightedEncoder } from '../encoders';
+import { ComposableStableEncoder } from '../encoders/composableStable';
+import { NestedJoinCallAttributes } from './types';
+import { replaceWrapped } from '../utils/replaceWrapped';
+import { bathcRelayerLibraryAbi } from '../../abi';
```
johngrantuk, line 7: small typo

src/entities/nestedExit/encodeCalls.ts
johngrantuk: I did see it was used in `getSender`?

src/entities/nestedExit/getQueryCallsAttributes.ts
johngrantuk: Think that helps a lot!

src/entities/nestedExit/getNestedExitCalls.ts
johngrantuk: I'm not 100% on this. I kind of feel like we should have a check/validation early in the code and throw if we are hitting any of the constraints. I'll make a note to discuss this during the stand up to get Daniels view.

src/entities/nestedExit/getNestedExitCalls.ts
johngrantuk: Thanks for clarifying. Would you add similar as a code comment. I think it'll help if we ever come back to this code in future.

src/entities/nestedExit/encodeCalls.ts
brunoguerios: Yeah, I added this comment before implementing `getSender`. It does help on that logic, so I'd say we should keep it 👍

src/data/providers/balancer-api/modules/nested-pool-state/index.ts
```diff
@@ -0,0 +1,265 @@
...
+    mapPoolType = (type: string): PoolType => {
+        switch (type) {
+            case 'WEIGHTED':
+                return PoolType.Weighted;
+            case 'PHANTOM_STABLE':
+                return PoolType.ComposableStable;
+            default:
+                throw new Error(`Unsupported pool type ${type}`);
+        }
+    };
```
brunoguerios, line 264: I'd like to highlight this method. I added it in order to separate pool type on API vs b-sdk. Alongside this, I changed `type` from `String` to `PoolType` within `NestedPool`. 
This is somewhat different from the pure join approach, because there we split `PoolState` and `PoolStateInput` to handle this difference on "internal" vs "external" objects.
Following pure join approach, I'd expect to have a `NestedPoolInput` that has `type` as `String` and only within the  `query` method I'd convert it to `NestedPool` with `type` as `PoolType`, but that seems like too much overhead for me 😬

package.json
agualis: @brunoguerios do we have an expected date for this?

brunoguerios: Not yet. It's currently being audited and after that it needs to go through governance approval. My guess is that it will be live in ~1 month.

PR #101, opened 2023-10-11
Replace dataQueries with multicallV3 (with batching)

johngrantuk:
* Remove BALANCER_POOL_DATA_QUERIES_ADDRESSES
* Replace old data queries call
* Remove tokenRates and dataQueries ABI
* Add multicallV3 ABI
* Update onChainPoolDataEnricher
* Removed queryFailed as not used anywhere and was misleading into thinking it was handled - logging error instead
* Delete old multicall implementations
* Fix scalingFactor scaling
* Add manual batching (instead of Viems) so we can maintain custom blockNo


src/data/onChainPoolDataViaReadContract.ts
```diff
@@ -0,0 +1,451 @@
...
+                    ...defaultCallsAux.parse(results, shift),
+                    ...stableCalls.parse(
+                        results,
+                        shift + defaultCallsAux.count,
+                    ),
+                }),
+            };
+        }
+        case 'GyroE':
+            if (poolTypeVersion === 2) {
```
brunoguerios, line 315: Why reverse the logic here? Doesn't it make more sense to assume that newer versions will follow poolTypeVersion 2 and use tokenRates?

.changeset/few-mayflies-change.md
```diff
@@ -0,0 +1,11 @@
+---
+"@balancer/sdk": minor
+---
+
+Replace dataQueries with multicallV3:
+
+- Add multicallV3 ABI
+- Add manual batching (instead of Viems) so we can maintain custom blockNo
```
brunoguerios, line 8: If I understood correctly, you're saying you're using multicallV3 directly because viem's multicall does not allow you to set a blockNumber - is that right?
Looking at the previous implementation, it seems that we're able to set the blockNumber, so I'm assuming I misunderstood something 🤔 
So, my question is: why aren't we using viem's multicall?

brunoguerios (commented):
Nice and clean implementation 👏 LGTM 👍

src/utils/constants.ts
```diff
@@ -112,14 +112,23 @@ export const STELLATE_URLS = {
...
-    [ChainId.ZKEVM]: '0xF24917fB88261a37Cc57F686eBC831a5c0B9fD39',
+export const MULTICALL: Record<number, Address> = {
+    [ChainId.ARBITRUM_ONE]: '0xca11bde05977b3631167028862be2a173976ca11',
+    [ChainId.AVALANCHE]: '0xca11bde05977b3631167028862be2a173976ca11',
+    [ChainId.GNOSIS_CHAIN]: '0xca11bde05977b3631167028862be2a173976ca11',
+    [ChainId.MAINNET]: '0xca11bde05977b3631167028862be2a173976ca11',
+    [ChainId.OPTIMISM]: '0xca11bde05977b3631167028862be2a173976ca11',
+    [ChainId.POLYGON]: '0xca11bde05977b3631167028862be2a173976ca11',
+    [ChainId.ZKEVM]: '0xca11bde05977b3631167028862be2a173976ca11',
+};
```
gmbronco, line 123: it's always the same, which is a feature, so we can just define it in the multicaller lib.

src/utils/constants.ts
```diff
@@ -175,3 +184,340 @@ export const DEFAULT_FUND_MANAGMENT = {
...
+    '0x32be2d0ddeaf3333501b24a28668ce373ba8e763000200000000000000000014',
+    '0x32f03464fdf909fdf3798f87ff3712b10c59bd86000000000000000000000005',
+    '0x4b718e0e2fea1da68b763cd50c446fba03ceb2ea00000000000000000000000b',
+    '0x68a69c596b3839023c0e08d09682314f582314e5000200000000000000000011',
+    '0x6f34a44fce1506352a171232163e7716dd073ade000200000000000000000015',
+    '0x9e2d87f904862671eb49cb358e74284762cc9f42000200000000000000000013',
+    '0xac4b72c01072a52b73ca71105504f1372efcce0d000000000000000000000003',
+    '0xbfd65c6160cfd638a85c645e6e6d8acac5dac935000000000000000000000004',
+    '0xe274c9deb6ed34cfe4130f8d0a8a948dea5bb28600000000000000000000000d',
+];
```
gmbronco, line 523: any way to filter them out in the subgraph, or RPC?

.changeset/few-mayflies-change.md
jxom: We do [allow you to set a block number](https://github.com/wagmi-dev/viem/blob/main/src/actions/public/multicall.ts#L112) FWIW. It's not documented, but we should document it 😅

src/data/onChainPoolDataViaReadContract.ts
johngrantuk: Yeah, that probably does make more sense although its worth double checking if any newer versions do come out. Changed.

src/utils/constants.ts
johngrantuk: These get used in the Subgraph provider query to filter. Or do you mean by default at source?

src/utils/constants.ts
johngrantuk: No longer needed now we're using Viems multicall which has config per chain.

src/utils/constants.ts
gmbronco: yes, at source, since we know the pool is bricked, we could set a flag somewhere that it's killed, is there something like that on a contract?

PR #106, opened 2023-10-16
Created Join for Composable Stable Pools;

lgahdl:


src/entities/join/poolJoin.ts
```diff
@@ -18,6 +19,7 @@ export class PoolJoin {
         const { customPoolJoins } = config || {};
         this.poolJoins = {
             Weighted: new WeightedJoin(),
+            PHANTOM_STABLE: new ComposableStableJoin(),
```
johngrantuk, line 22: Does `PHANTOM_STABLE` cover all stable pool types?

src/entities/join/types.ts
```diff
@@ -54,6 +54,7 @@ export type JoinQueryResult = {
     amountsIn: TokenAmount[];
     fromInternalBalance: boolean;
     tokenInIndex?: number;
+    bptIndex?:number;
```
johngrantuk, line 57: Lets create a `BaseJoinQuery` type and then separate query types for each pool type, e.g. `WeightedQueryResult`, `StableQueryResult`. Then `JoinQueryResult` can be a union of those. That way we know what properties belong to what pool type. e.g. we know `bptIndex` belongs to stable pools.

src/entities/join/composable-stable/validateInputs.ts
johngrantuk, line 1: This is just the same code as Weighted so lets move to a common location and reuse.

src/entities/join/poolJoin.ts
lgahdl: Only Composable Stables

johngrantuk: Can you add a comment that explains that here, just as a reminder for future.

test/composableStableJoin.integration.test.ts
```diff
@@ -0,0 +1,387 @@
...
+        expect(queryResult.bptOut.amount).to.deep.eq(bptOut.amount);
+
+        // Expect all assets to have a value for amount in
+        expect(queryResult.tokenInIndex).toBeDefined;
+        queryResult.amountsIn
+            .filter((_, index) => index !== bptIndex)
+            .forEach((a) => {
+                expect(a.amount > BigInt(0)).to.be.true;
+            });
+        expect(queryResult.tokenInIndex).toBeUndefined;
```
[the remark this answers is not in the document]
johngrantuk: Strange! Seems like `expect(queryResult.tokenInIndex === undefined).toEqual(true or false);` works.
We also have this in weightedJoin.integration.tests.ts and should also change any use of `toBeDefined` and `toBeUndefined`, this can be done in a separate PR and I'll make an Asana ticket.

test/composableStableJoin.integration.test.ts
```diff
@@ -0,0 +1,387 @@
...
+                ...txInput.poolInput.tokens.map((t) =>
+                    parseUnits('100', t.decimals),
+                ),
+            ],
+            process.env.ETHEREUM_RPC_URL as string,
+            blockNumber,
+        );
+    });
+
+    test('unbalanced join', async () => {
```
brunoguerios, line 103: Maybe an opportunity to reuse code could come from identifying patterns between `unbalanced join` tests on different types of pools (e.g. weighted, composable stable, etc..).
I feel like the assertions should be pretty similar between them. If that's true, we could have something like a shared `testUnbalancedJoin` between pool types.
Likely not something for this PR, but for a separate one.

brunoguerios (commented):
Left some minor comments, but otherwise LGTM 👍

test/composableStableJoin.integration.test.ts
```diff
@@ -0,0 +1,371 @@
...
+
+    // Confirm slippage - only to amount in not bpt out
+    const expectedMaxAmountsIn = queryResult.amountsIn.map((a) =>
+      txInput.slippage.applyTo(a.amount),
+    );
+    expect(expectedMaxAmountsIn).to.deep.eq(maxAmountsIn);
+    expect(minBptOut).to.eq(bptOut.amount);
+  });
+
+  async function doTransaction(txIp: TxInput) {
```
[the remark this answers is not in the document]
lgahdl: It's not the same actually, the "expectedDeltas" variable is different for the composable stable, because the bpt amount is duplicated

test/composableStableJoin.integration.test.ts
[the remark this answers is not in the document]
lgahdl: True, That's very odd, the tests shouldn't be working

test/composableStableJoin.integration.test.ts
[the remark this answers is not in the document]
johngrantuk: Ahh sorry, didn't notice that. As mentioned in the other PR we can probably still reuse a lot across tests.

tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-    "include": ["src"],
+    "include": ["src", "test"],
```
johngrantuk, line 2: Unless there's a strong reason I don't think we should include test here.

test/composableStableJoin.integration.test.ts
```diff
@@ -0,0 +1,311 @@
...
+
+        const { queryResult, maxAmountsIn, minBptOut, value } = await doJoin({
+            ...txInput,
+            joinInput,
+        });
+        // Query should use same bpt out as user sets
+        expect(queryResult.bptOut.amount).to.deep.eq(bptOut.amount);
+
+        // We only expect single asset to have a value for amount in
+        expect(queryResult.tokenInIndex).toBeDefined;
```
johngrantuk, line 214: Because of the issue we saw before testing against undefined lets use:
`expect(checkedValue === undefined).toEqual(true);`

test/composableStableJoin.integration.test.ts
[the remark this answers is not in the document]
johngrantuk: I've already done the weighted tests (you'll need to merge latest join) but we still need to do it here @lgahdl .

test/composableStableJoin.integration.test.ts
johngrantuk: Created an Asana for this.

test/lib/utils/joinHelper.ts
```diff
@@ -0,0 +1,74 @@
+import { replaceWrapped, Token } from '../../../src';
+import { sendTransactionGetBalances } from './helper';
+import { expect } from 'vitest';
+import { JoinTxInput } from './types';
+
+export const doJoin = async (txInput: JoinTxInput) => {
```
johngrantuk, line 6: Nice to reuse this 👏
Could you add a bit of documentation to the function that explains what its doing/testing so we don't have to read through it all when we need a reminder.

test/lib/utils/joinHelper.ts
```diff
@@ -0,0 +1,74 @@
...
+        (t) => new Token(chainId, t.address, t.decimals),
+    );
+    const bptIndex = poolInput.tokens.findIndex(
+        (t) => t.address === poolInput.address,
+    );
+    let tokensForBalanceCheck; // Replace with native asset if required
+    const poolTokensAddr = checkNativeBalance
+        ? replaceWrapped(poolTokens, chainId).map((t) => t.address)
+        : poolTokens.map((t) => t.address);
+    if (bptIndex >= 0) {
```
johngrantuk, line 35: Also a bit of comment here and LN53 would be useful as an explainer of why this is being done.

tsconfig.json
lgahdl: I Included "test" in here to get rid of TS Errors on my test files, but it's not a big deal, I can remove it

test/composableStableJoin.integration.test.ts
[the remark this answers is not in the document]
lgahdl: We can also use "expect(value).to.be.undefined" or "expect(value).to.not.be.undefined", works fine

johngrantuk (changes requested):
Still got the tsconfig change to revert. 
And seems like there's still an issue with your formatting/linter. When I right-click format document on, e.g. `test/composableStableJoin.integration.test.ts` it fixes some issues. This happens automatically for me when I edit/save normally.

src/entities/join/composable-stable/composableStableJoin.ts
```diff
@@ -0,0 +1,215 @@
...
+                        input.amountsIn,
+                        BigInt(0),
+                    ),
+                    tokenInIndex: undefined,
+                };
+            }
+            case JoinKind.SingleAsset: {
+                if (bptIndex === undefined) {
+                    throw new Error('bptIndex is necessary');
+                }
```
brunoguerios, line 136: I feel like these checks are better suited for a `validateInputs` method that should run before anything else. What do you think?

src/entities/join/composable-stable/composableStableJoin.ts
```diff
@@ -0,0 +1,215 @@
...
+            (t) => t.address === poolState.address,
+        );
+        const amounts = this.getAmountsQuery(poolState.tokens, input, bptIndex);
+        const amountsWithoutBpt = {
+            ...amounts,
+            maxAmountsIn: [
+                ...amounts.maxAmountsIn.slice(0, bptIndex),
+                ...amounts.maxAmountsIn.slice(bptIndex + 1),
+            ],
+        };
```
brunoguerios, line 34: How about we make `getAmountsQuery` output both `amounts` and `amountsWithoutBpt`?
I think the main function (i.e. query) would be cleaner and easier to read if we moved this type of manipulation to inner functions.

test/composableStableJoin.integration.test.ts
```diff
@@ -0,0 +1,309 @@
...
+        const bptIndex = txInput.poolInput.tokens.findIndex(
+            (t) => t.address === txInput.poolInput.address,
+        );
+        const poolTokensWithoutBpt = txInput.poolInput.tokens
+            .map((t) => new Token(chainId, t.address, t.decimals))
+            .filter((_, index) => index !== bptIndex);
+
+        const amountsIn = poolTokensWithoutBpt.map((t) =>
+            TokenAmount.fromHumanAmount(t, '1'),
+        );
```
brunoguerios, line 98: Should we go ahead and update `amountsIn` to minimal token info (address + rawAmount)? Or is this better suited for a separate refactor PR that will cover all regular joins? @johngrantuk

test/composableStableJoin.integration.test.ts
```diff
@@ -0,0 +1,309 @@
...
+
+        // Confirm slippage - only bpt out
+        const expectedMinBpt = txInput.slippage.removeFrom(
+            queryResult.bptOut.amount,
+        );
+        expect(expectedMinBpt).to.deep.eq(minBptOut);
+        const expectedMaxAmountsIn = amountsIn.map((a) => a.amount);
+        expect(expectedMaxAmountsIn).to.deep.eq(
+            maxAmountsIn.filter((_, index) => index !== bptIndex),
+        );
```
brunoguerios, line 132: I see that we have a few test assertions within `doJoin` and some other spread between tests. How similar are they between tests? Does it make sense to extract all checks to a separate `assertResults` function?
This way we could separate join functionality (`doJoin`) and test assertions (`assertResults`).

src/entities/join/composable-stable/composableStableJoin.ts
johngrantuk: Yeah, I was thinking this can be done in a later refactor after this is merged. I've made an Asana already.

tsconfig.json
johngrantuk: My bad - Alberto confirmed this is ok [here](https://github.com/balancer/b-sdk/pull/111#discussion_r1370192599)

src/entities/join/composable-stable/composableStableJoin.ts
lgahdl: Actually, this bptIndex will not be undefined in any case, this "if" was necessary before I adapted the types, I will remove this, but thanks

test/composableStableJoin.integration.test.ts
lgahdl: Yeah, maybe I can move these assertions back to the test file rather than in a helper function, wdyt?

brunoguerios: Yeah, not only having them all within the test files, but reusing them would be great. A helper function within the test file seems to be a good solution to achieve that.

src/entities/join/composable-stable/composableStableJoin.ts
brunoguerios: You'll still have to check for bptIndex being different than -1 somewhere, right?

lgahdl: Right, that's true

test/composableStableJoin.integration.test.ts
johngrantuk: Yeah, I was thinking this can be done in a later refactor after this is merged. I've made an Asana already.

test/composableStableJoin.integration.test.ts
johngrantuk: @brunoguerios - I've done a refactor of the tests, take a look and let me know what you think.

test/composableStableJoin.integration.test.ts
```diff
@@ -0,0 +1,283 @@
...
+            const joinInput = {
+                ...input,
+                amountsIn,
+                useNativeAssetAsWrappedAmountIn: true,
+            };
+            const joinResult = await doJoin({
+                ...txInput,
+                joinInput: {
+                    ...joinInput,
+                    useNativeAssetAsWrappedAmountIn: true,
```
brunoguerios, line 139: You should be able to remove this, since you're already adding it to joinInput above.

test/composableStableJoin.integration.test.ts
```diff
@@ -0,0 +1,283 @@
...
+                },
+            });
+            assertUnbalancedJoin(
+                txInput.client.chain?.id as number,
+                txInput.poolStateInput,
+                { ...joinInput, useNativeAssetAsWrappedAmountIn: true },
```
brunoguerios, line 145: Same here.

test/composableStableJoin.integration.test.ts
```diff
@@ -0,0 +1,283 @@
...
+            );
+        });
+
+        test('with native', async () => {
+            const joinResult = await doJoin({
+                ...txInput,
+                joinInput: {
+                    ...joinInput,
+                    useNativeAssetAsWrappedAmountIn: true,
+                },
```
brunoguerios, line 186: If you redeclare `joinInput` with this flag before `joinResult`, you should be able to reuse as you did on the previous test.

test/composableStableJoin.integration.test.ts
```diff
@@ -0,0 +1,283 @@
...
+                txInput.slippage,
+            );
+        });
+        test('with native', async () => {
+            const joinResult = await doJoin({
+                ...txInput,
+                joinInput: {
+                    ...joinInput,
+                    useNativeAssetAsWrappedAmountIn: true,
+                },
```
brunoguerios, line 233: Same here about redeclaring and reusing below.

test/lib/utils/joinHelper.ts
```diff
@@ -0,0 +1,401 @@
...
+export function assertSingleToken(
+    chainId: ChainId,
+    poolStateInput: PoolStateInput,
+    joinInput: SingleAssetJoinInput,
+    joinResult: JoinResult,
+    slippage: Slippage,
+) {
+    const { txOutput, joinQueryResult, joinBuildOutput } = joinResult;
+
+    if (!joinQueryResult.tokenInIndex) throw Error('No index');
```
brunoguerios, line 219: Maybe change this to an assertion where `tokenInIndex` should be defined (and even >= 0 to make sure it's not -1)?
If you think it should indeed be throwing like this, I'd update it to check for `!== undefined` because the current check will throw for index `0`.

brunoguerios (approved):
Wow! This looks quite complex at a first glimpse 😅 , but I think it will be very beneficial on the long run. With a robust set of pre-defined assertions, adding new tests should be very simple/straightforward.
Thanks for the refactor. 🙏 
Ps: I left a few minor suggestions, but overall looks good to go 👍

test/lib/utils/joinHelper.ts
johngrantuk: This was actually just to narrow the type rather than testing anything but definitely need to keep the 0 index in mind so changed it to check for undefined. Thanks.

PR #107, opened 2023-10-17
Nested Exit

brunoguerios:
- Add proportional nested exit
- Add single token nested exit

### DIFF
--- a/src/entities/nestedExit/getNestedExitCalls.ts
+++ b/src/entities/nestedExit/getNestedExitCalls.ts
@@ -1,28 +1,93 @@
 import { Token } from '../token';
 import { NestedExitInput, NestedExitCall } from './types';
-import { NestedPoolState } from '../types';
+import { NestedPool, NestedPoolState } from '../types';
 import { TokenAmount } from '../tokenAmount';
+import { Address } from '../../types';
 
 export const getNestedExitCalls = (
     {
         bptAmountIn,
         chainId,
         accountAddress,
+        tokenOut,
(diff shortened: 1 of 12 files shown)


src/entities/nestedExit/getNestedExitCalls.ts
```diff
@@ -0,0 +1,81 @@
...
+            toInternalBalance,
+            // TODO: previous implementation of nested exit didn't add an outputReferenceKey for the BPT token,
+            // but if I remove it from here, peek logic fails. Need to investigate why.
+            // Once we figure this out, we should be able to replace sortedTokens by sortedTokensWithoutBpt
+            outputReferenceKeys: sortedTokens.map(
+                (token) =>
+                    100n +
+                    BigInt(poolsSortedByLevel.indexOf(pool)) * 10n +
+                    BigInt(sortedTokens.indexOf(token)),
+            ),
```
brunoguerios, line 74: @johngrantuk - I'd like to highlight this part of the code here - at the end I think we'll need input from the SC team, but I'd like to run by someone internally first, to make sure there's nothing obvious that I'm missing regarding indexes, etc..
Basically it seems that relayerQueries ignores the outputReference `index` I set and query fails if I remove the outputReference related to the BPT token, because it reads the incorrect amount.
On the other hand, having an outputReference related to the BPT token breaks the transaction itself.
To fix this, I first add it to the query and later remove when building the transaction.
It works, but I'd expect this kind of manipulation to not be necessary between query vs actual transaction.

PR #108, opened 2023-10-18
Composable stable exit

lgahdl:


src/entities/exit/types.ts
johngrantuk, line 1: Could you do a similar refactor as we did on [Join types](https://github.com/balancer/b-sdk/pull/106/commits/6a08b5fbbb36e98e57bcbf02c903feba6617430c).

test/composableStableExit.integration.test.ts
```diff
@@ -0,0 +1,336 @@
...
+      chainId,
+      rpcUrl,
+      bptIn,
+      kind: ExitKind.PROPORTIONAL,
+      exitWithNativeAsset: true,
+    };
+
+    // Note - checking native balance
+    const { queryResult, maxBptIn, minAmountsOut } = await doTransaction({
+      ...txInput,
```
johngrantuk, line 221: Similar to join lets reuse any similar helper functions across tests.

src/entities/exit/poolExit.ts
```diff
@@ -18,6 +19,7 @@ export class PoolExit {
         const { customPoolExits } = config || {};
         this.poolExits = {
             Weighted: new WeightedExit(),
+            PHANTOM_STABLE: new ComposableStableExit(),
```
johngrantuk, line 22: Can you add a comment (same as Join) that explains the `PHANTOM_STABLE` naming as its a bit confusing and useful for reference.

test/composableStableExit.integration.test.ts
lgahdl: But the composable doTransaction has a different expectedAmounts than the weighted, maybe I could use the same function, but adapt for each pool type?

johngrantuk: I think there's a lot of similar functionality that could be reused?

src/entities/exit/composable-stable/composableStableExit.ts
```diff
@@ -0,0 +1,200 @@
...
+            tokenOutIndex: amounts.tokenOutIndex,
+            toInternalBalance: !!input.toInternalBalance,
+            bptIndex,
+        };
+    }
+
+    private getAmountsQuery(
+        tokens: Token[],
+        input: ExitInput,
+        bptIndex?: number,
```
johngrantuk, line 82: We can make this non optional.

test/lib/utils/exitHelper.ts
```diff
@@ -0,0 +1,348 @@
...
+        recipient: testAddress,
+    });
+
+    return {
+        exitBuildOutput,
+        exitQueryResult,
+    };
+};
+
+function getTokens(poolStateInput: PoolStateInput): Address[] {
```
johngrantuk, line 54: we can reuse/split out from joinHelper?

test/lib/utils/exitHelper.ts
```diff
@@ -0,0 +1,348 @@
...
+    // if exactIn slippage cannot be applied to bptIn, else should use bptIn with slippage applied
+    const maxBptIn = isExactIn
+        ? exitQueryResult.bptIn.amount
+        : slippage.applyTo(exitQueryResult.bptIn.amount);
+
+    const expectedBuildOutput: Omit<ExitBuildOutput, 'call'> = {
+        minAmountsOut,
+        maxBptIn,
+        to: BALANCER_VAULT,
+        // Value should equal value of any wrapped asset if using native
```
johngrantuk, line 341:
```suggestion
        // Value should always be 0 for exits
```

johngrantuk (approved):
Looking good! Can go ahead and merge once the last little recent suggestions are updated.

test/lib/utils/exitHelper.ts
lgahdl: Yeah, it's the same function, I made it reusable, thanks!

PR #109, opened 2023-10-18
Viem anvil

agualis:
Setup @viem/anvil to automatically run anvil fork with default parameters.


.github/workflows/checks.yml
```diff
@@ -43,13 +43,11 @@ jobs:
         uses: ./.github/actions/setup
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
-      - name: Run Anvil Node
-        run: anvil --fork-url ${{ secrets.ETHEREUM_RPC_URL }} &
       - name: Test
         run: pnpm test:ci
         env:
-          ETHEREUM_RPC_URL: ${{ secrets.ETHEREUM_RPC_URL }}
```
johngrantuk, line 51: These RPCs will likely be used for example/debug scripts too so unless there's a strong reason can we keep them without the `VITE_` prefix?

brunoguerios:
Awesome work @agualis ! 👏 
I do have a few questions 😅 
How does this actually work? Does the fork automatically runs before each test? Or does it run only once before all tests?
I ask because while implementing nested joins/exits with relayer queries, I need the fork to not be reset before each test
I know it’s an edge case, but I’d like to get a sense on how to handle this.

.github/workflows/checks.yml
agualis: Fixed by using `import { loadEnv } from 'vite';` in `vitest.config.ts`

PR #110, opened 2023-10-18
Add node >= 18 to engines in package.json

agualis:
Viem requires Node 18+: https://viem.sh/docs/compatibility.html#platform-compatibility

We state that in engines section of package.json


johngrantuk (approved):

johngrantuk:
Useful to note this mainly serves as a warning/doc as described here: https://www.marcusoft.net/2015/03/packagejson-and-engines-and-enginestrict.html which can be useful.

PR #111, opened 2023-10-24
Use vitest globals setup

agualis:
Adds [globals vitest setup](https://vitest.dev/config/#globals) so that test files do not need to explicitly import `describe, it, beforeEach` etc

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,5 +1,5 @@
 {
-    "include": ["src"],
+    "include": ["src", "test"],
     "compilerOptions": {
         "target": "ESNext",
         "module": "ESNext",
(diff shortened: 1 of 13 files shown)


tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-    "include": ["src"],
+    "include": ["src", "test"],
```
brunoguerios, line 2: Is there any side effects to including `test` over here?

agualis: Tests files now will be typechecked. 
The build won't be affected as the entry point for `tsup` (`src/index.ts` by default) and the build graph does not reference `/tests` code.

johngrantuk (approved):

agualis:
@johngrantuk Should I merge or do you prefer to wait for something?

PR #112, opened 2023-10-31
Test refactor to use local forks

johngrantuk:
* Update anvil setup to work with multiple networks
* Add Fantom and Polygon forks
* Update all tests that were using RPC directly to run against local fork using mock pool data (to avoid slow calls to SG)
* Split sor.test.ts into separate Weighted and ComposableStable tests
* Re-org, tidy and consistent naming for pool test files


test/anvil/anvil-global-setup.ts
```diff
@@ -1,24 +1,71 @@
...
+        forkBlockNumber: 65313450n,
+    },
+];
+
 export default async function () {
+    for (const network of NETWORKS) {
+        const config = getAnvilOptions(network);
+        await startFork(config);
+    }
+}
```
brunoguerios, line 42: How does this work? Does it start all forks before all tests only once? Or does it restart on every test? I'm trying to get a sense of the impact of this change.
Maybe a better question is: how did this affect test performance? Is it faster/slower than what we had before?

johngrantuk: It only starts each network once for all tests.
The PR significantly improves test performance for two reasons, 1. Smaller list of pools to fetch onchain state 2. Faster to fetch onchain state locally than over rpc.
There's still a bottleneck with the Subgraph tests.

test/lib/utils/helper.ts
```diff
@@ -255,7 +255,6 @@ export async function findTokenBalanceSlot(
  * @param slots Slot that stores token balance in memory - use npm package `slot20` to identify which slot to provide
  * @param balances Balances in EVM amounts
  * @param jsonRpcUrl Url with remote node to be forked locally
- * @param blockNumber Number of the block that the fork will happen
```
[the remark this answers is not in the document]
johngrantuk: Yeah, it was removed in the earlier [PR](https://github.com/balancer/b-sdk/pull/109/files). Its not needed right now and would be easy to add if do.

test/anvil/anvil-global-setup.ts
```diff
@@ -1,24 +1,71 @@
...
+
+function getAnvilOptions(network: NetworkSetup): CreateAnvilOptions {
     let forkUrl: string;
-    if (process.env.ETHEREUM_RPC_URL) {
-        forkUrl = process.env.ETHEREUM_RPC_URL;
+    if (process.env[network.rpcEnv]) {
```
lgahdl, line 46: This "if" statement works if the variable is available on the .env file, but is empty, but does not work if the variable is not available in the .env file at all(process.env[network.rpcEnv] is a string 'undefined' in this case), so it should be`process.env[network.rpcEnv] !== 'undefined' && process.env[network.rpcEnv]`

test/anvil/anvil-global-setup.ts
```diff
@@ -1,24 +1,71 @@
...
+type NetworkSetup = {
+    rpcEnv: string;
+    fallBackRpc: string | undefined;
+    port: number;
+    forkBlockNumber: bigint;
+};
+
+type Networks = NetworkSetup[];
+
+const NETWORKS: Networks = [
```
agualis, line 13:
```suggestion
type NetworksWithFork = Extract<
    keyof typeof ChainId,
    'MAINNET' | 'POLYGON' | 'FANTOM'
>;

const NETWORKS: Record<NetworksWithFork, NetworkSetup> = {
    MAINNET: {
        rpcEnv: 'ETHEREUM_RPC_URL',
        fallBackRpc: 'https://cloudflare-eth.com',
        port: 8545,
        forkBlockNumber: 18043296n,
    },
    POLYGON: {
        rpcEnv: 'POLYGON_RPC_URL',
        // Public Polygon RPCs are usually unreliable
        fallBackRpc: undefined,
        port: 8137,
        // Note - this has to be >= highest blockNo used in tests
        forkBlockNumber: 44215395n,
    },
    FANTOM: {
        rpcEnv: 'FANTOM_RPC_URL',
        // Public Fantom RPCs are usually unreliable
        fallBackRpc: undefined,
        port: 8138,
        forkBlockNumber: 65313450n,
    },
};

export default async function () {
    for (const network of Object.values(NETWORKS)) {
        const config = getAnvilOptions(network);
        await startFork(config);
    }
}
```

agualis: Instead of an array I would use an object that we can reference in other tests (see comments below)

test/composableStable.integration.test.ts
```diff
@@ -0,0 +1,107 @@
...
+import { SwapKind, SwapOptions } from "../src/types";
+import { BasePool } from "../src/entities/pools";
+import { MockPoolProvider } from "./lib/utils/mockPoolProvider";
+
+import testPools from "./lib/testData/testPools/composableStable_17473810.json";
+import { RawStablePool } from "../src";
+
+describe("ComposableStable Swap tests", () => {
+  const chainId = ChainId.MAINNET;
+  const rpcUrl = "http://127.0.0.1:8545";
```
agualis, line 20: If you refactor NETWORKS to an object, here you can replace rpcUrl by: 
`NETWORKS.MAINNET.rpcUrl` (I would rename NETWORKS to ANVIL_NETWORKS or alike)

test/fantom.test.ts
```diff
@@ -0,0 +1,103 @@
...
+
+import testPools from "./lib/testData/testPools/fantom_65313450.json";
+import { RawBasePool } from "../src";
+
+describe(
+  "Fantom SOR",
+  () => {
+    const chainId = ChainId.FANTOM;
+    const inputToken = NATIVE_ASSETS[chainId];
+    const rpcUrl = "http://127.0.0.1:8138/";
```
agualis, line 23: This would be `ANVIL_NETWORKS.FANTOM.rpcUrl`

test/fxPool.integration.test.ts
```diff
@@ -7,18 +7,21 @@ import { BATCHSIZE, ChainId, VAULT } from '../src/utils';
...
     sorGetSwapsWithPools,
 } from '../src';
+import { MockPoolProvider } from './lib/utils/mockPoolProvider';
+
+import testPools from './lib/testData/testPools/fx_43667355.json';
 
 describe('fx integration tests', () => {
     const chainId = ChainId.POLYGON;
-    const rpcUrl = process.env.POLYGON_RPC_URL || 'https://polygon-rpc.com';
+    const rpcUrl = 'http://127.0.0.1:8137/';
```
agualis, line 24: This would be ANVIL_NETWORKS.POLYGON.rpcUrl

test/gyro3Pool.integration.test.ts
```diff
@@ -18,7 +18,7 @@ import {
 
 describe('gyro3 integration tests', () => {
     const chainId = ChainId.POLYGON;
-    const rpcUrl = process.env['POLYGON_RPC_URL'] || 'https://polygon-rpc.com';
+    const rpcUrl = 'http://127.0.0.1:8137/';
```
agualis, line 21: This would be ANVIL_NETWORKS.POLYGON.rpcUrl

test/gyroEV2Pool.integration.test.ts
```diff
@@ -21,7 +21,7 @@ import { MockPoolProvider } from './lib/utils/mockPoolProvider';
 
 describe('gyroEV2: WMATIC-stMATIC integration tests', () => {
     const chainId = ChainId.POLYGON;
-    const rpcUrl = process.env.POLYGON_RPC_URL || 'https://polygon-rpc.com';
+    const rpcUrl = 'http://127.0.0.1:8137/';
```
agualis, line 24: This would be ANVIL_NETWORKS.POLYGON.rpcUrl

test/weightedPool.integration.test.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    const BAL = new Token(
+      chainId,
+      "0xba100000625a3754423978a60c9317c58a424e3D",
+      18,
+      "BAL"
+    );
+
+    const swapOptions: SwapOptions = {
+      block: 17473810n,
```
agualis, line 48: Love that the name of the json contains the block. Clever idea! 💡

agualis (changes requested):
Nice PR 🎉 
Left one suggestion.

test/anvil/anvil-global-setup.ts
johngrantuk: Thanks for pointing that out!

PR #113, opened 2023-11-01
Constraint validations

johngrantuk:
General rules:
  * Can only join/exit with the main tokens
  * Main tokens only supported to a max of 1 level of nesting
  * Can still pure join with > 1 level 
  * A main token can't be a token in > 1 pool

We can only do minimal validation without introducing complexity/data overhead that the API should be handling. 
e.g. We can't validate pool state without constructing it which would require more complexity and data.
So we assume pool state is correct.

### DIFF
--- a/src/entities/utils/constraintValidation.ts
+++ b/src/entities/utils/constraintValidation.ts
@@ -14,7 +14,9 @@ export function constraintValidation(
     e.g. We can't validate pool state without constructing it which would require more complexity and data.
     So we assume pool state is correct. 
     */
-    const topLevel = nestedPoolState.pools[0].level;
+    // pools may not be in order so find highest level which will be top pool
+    const topLevel = Math.max(...nestedPoolState.pools.map((p) => p.level));
+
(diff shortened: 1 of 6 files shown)


src/entities/utils/constraintValidation.ts
```diff
@@ -0,0 +1,35 @@
...
+      * Can only join/exit with the main tokens
+      * Main tokens only supported to a max of 1 level of nesting
+      * Can still pure join with > 1 level 
+      * A main token can't be a token in > 1 pool
+
+    We can only do minimal validation without introducing complexity/data overhead that the API should be handling. 
+    e.g. We can't validate pool state without constructing it which would require more complexity and data.
+    So we assume pool state is correct. 
+    */
+    const topLevel = nestedPoolState.pools[0].level;
```
brunoguerios, line 17: I was under the impression that since the API won't guarantee token sorting within the pool, it also wouldn't guarantee pool sorting within nested pools.
If that's true, we should sort pools to make sure they are ordered by level.

brunoguerios (approved):

PR #117, opened 2023-11-03
setup vitest/anvil to allow parallel threads

agualis:
It extends the existing #112 refactor PR to allow running tests in parallel  threads (when the running machine has multicore).

The new setup enforces integration tests to explicitly call 
`const { rpcUrl } = await startFork(ANVIL_NETWORKS.<NETWORK>);`

to have more flexibility (as in the future, we could decide to reset fork state in `afterEach/afterAll beforeEach/beforeAll` when needed).

In Github Actions CI, (no threads available) the tests will run sequentially (we can still split them in parallel jobs if we need faster feedback).

In local machines with multicore, the tests will run in parallel, meaning that we will have much better performance: 

parallel-tests


johngrantuk (approved):
This does seem super fast! Thanks!

PR #118, opened 2023-11-05
Refactoring all exit tests, exit enums and parameters names;

lgahdl:


johngrantuk (approved):
Think its fine to merge this (once lint/tests are fixed) then we can do a final review off the main `composable-stable-exit` branch.

PR #119, opened 2023-11-06
Refactor Input amounts

johngrantuk:
Currently join inputs have TokenAmount as amountsIn type, but this adds too much complexity for the user to properly build the input. This task aims to replace that with a struct with more basic types:
`
InputAmount: {
    address: Address;
    decimals: number;
    rawAmount: bigint;
}
`

### DIFF
--- a/examples/join/weighted.ts
+++ b/examples/join/weighted.ts
@@ -1,3 +1,4 @@
+// Run with - pnpm example ./examples/join/weighted.ts
 import { config } from 'dotenv';
 config();
 
@@ -9,8 +10,6 @@ import {
     PoolJoin,
     PoolStateInput,
     Slippage,
-    Token,
-    TokenAmount,
     UnbalancedJoinInput,
 } from '../../src';
 import {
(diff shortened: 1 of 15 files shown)


johngrantuk:
@lgahdl - note you will have to update the CS Exit PR to match the same type (once this PR is approved/merged)

examples/join/weighted.ts
```diff
@@ -64,12 +63,12 @@ const join = async () => {
...
-        (t) => new Token(chainId, t.address, t.decimals),
-    );
-    const amountsIn = poolTokens.map((t) =>
-        TokenAmount.fromHumanAmount(t, '1'),
-    );
+
+    const amountsIn = poolState.tokens.map((t) => ({
+        rawAmount: parseUnits('1', t.decimals),
+        decimals: t.decimals,
+        address: t.address,
```
agualis, line 70: I personally like the `TokenAmount` [value object](https://martinfowler.com/bliki/ValueObject.html) abstraction and I think that using it (or a similar abstraction) in the internals of the frontend could be useful. (if you already have a token instance, having methods such as `fromHumanAmount` make the code easier to read).

That said, from an API consumer point of view, I think that it is better to leave more flexibility and freedom to the user and I agree that allowing primitives is better.

agualis (approved):

examples/join/weighted.ts
johngrantuk: It might be useful to add a `toInputAmount` to the TokenAmount?

agualis: I think it could be handy. Maybe also for you to use it in the sdk internals? 🤔

lgahdl:
> @lgahdl - note you will have to update the CS Exit PR to match the same type (once this PR is approved/merged)

done

PR #122, opened 2023-11-09
Manually add BPT in CS API Provider

johngrantuk:
We are working on assumption that API will return BPT token in token list (as current SG does) It does not currently do this so we have to add it manually.

### DIFF
--- a/src/data/providers/balancer-api/modules/pool-state/index.ts
+++ b/src/data/providers/balancer-api/modules/pool-state/index.ts
@@ -78,14 +78,33 @@ export class Pools {
     constructor(private readonly balancerApiClient: BalancerApiClient) {}
 
     async fetchPoolState(id: string): Promise<PoolStateInput> {
-        const {
-            data: { poolGetPool },
-        } = await this.balancerApiClient.fetch({
+        const { data } = await this.balancerApiClient.fetch({
             query: this.poolStateQuery,
             variables: {
                 id,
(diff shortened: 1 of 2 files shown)


src/data/providers/balancer-api/modules/pool-state/index.ts
```diff
@@ -78,14 +78,33 @@ export class Pools {
...
             },
         });
+        const poolGetPool: PoolStateInput = data.poolGetPool;
+        /**
+         * TODO:
+         * We are working on assumption that API will return BPT token in token list (as current SG does)
+         * It does not currently do this so we have to add it manually
+         */
+        if (poolGetPool.type === 'PHANTOM_STABLE') {
+            let missingBPTIndex = 0;
```
brunoguerios, line 94: Since they are still experimenting with adding/removing the BPT on v3-canary branch, we could make this code resilient by simply checking if pool tokens contains the pool address. If yes, then BPT is present and we don't need to add it.

brunoguerios (approved):
I added a small suggestion, but not something critical if we are able to sync with the API team on updates on their side. Otherwise LGTM!

src/data/providers/balancer-api/modules/pool-state/index.ts
johngrantuk: Good suggestion! Added that check now.

PR #124, opened 2023-11-10
Refactor Join/Exit Interface to Add/Remove Liquidity

brunoguerios:
Main changes:
- rename Join to AddLiquidity (and all its variations)
- rename Exit to RemoveLiquidity (and all its variations)

Other changes:
- rename several "Result" to "Output" to keep consistency with "Input"
- rename SingleAsset to SingleToken to follow standard defined on [Notion](https://www.notion.so/Interfaces-eeb0b72721234b1198813c8b7ad6f102)

Notes:
- did not change external interfaces (e.g. encoders from v2 are still using join/exit)
- used AddLiquidity always as prefix for consistency (e.g. AddLiquidityUnbalanced instead of UnbalancedJoin)
- used RemoveLiquidity always as prefix for consistency (e.g. RemoveLiquidityUnbalanced instead of UnbalancedExit)

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1205728223153979


johngrantuk (approved):
A lot of changes! But LGTM.

PR #128, opened 2023-11-15
New version scaffold

johngrantuk:
AddLiquidity:
* Add balancerVersion field to PoolStateInput (API will need to implement this)
    * Examples currently fail because API doesn't provide this yet
* Each pool type will have addLiquidityV2 and addLiquidityV3 where relevant encoding/query is implemented

SOR:
* Assuming maths and path selection stays the same.
* Added balancerVersion parameter to SorConfig
* Data providers/enrichers can map from V3 API/onchain format to existing data interfaces:
  * e.g. subgraphPoolProviderV2/V3 and onChainPoolDataEnricherV2/V3
  * SmartOrderRouter class checks balancerVersion in config and instantiates correct providers
* Output:
  * SwapV2/V3 map to relevant output interface for tx and query
  * check balancerVersion
  * getSwaps & sorGetSwapsWithPools return SwapV2 | SwapV3

TODO:
* Update Composable (Gyro are covered by weighted)
* Remove Liquidity
* Nested joins/exits
* Added various TODOs as place holders for future updates, uses `TODO V3` flags


src/entities/addLiquidity/weighted/addLiquidityWeightedV3.ts
```diff
@@ -0,0 +1,168 @@
...
+    AddLiquidityWeightedV3Call,
+} from '../types';
+import { AddLiquidityAmounts, PoolState } from '../../types';
+import {
+    doAddLiquidityQueryV3,
+    getAmounts,
+    parseAddLiquidityArgsV3,
+} from '../../utils';
+
+export class AddLiquidityWeightedV3 implements AddLiquidityBase {
```
danielmkm, line 19: With the invariant approach in v3, we won't need this distinction anymore. We can have a generalized `AddLiquidityV3` that can cover all pool types (minus custom scenarios).

brunoguerios:
Close as duplicate from #180

PR #131, opened 2023-11-21
Weighted Pool Creation

lgahdl:
- Created CreatePool class to create any type of pool;
- Test for Weighted Pool Creation with helpers;
- Utils functions for tests;


src/entities/createPool/createPool.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    constructor() {
+        this.createPoolTypes = {
+            WEIGHTED: new CreatePoolWeighted(),
+        };
+    }
+
+    private getCreatePool(poolType: string): CreatePoolBase {
+        if (!this.createPoolTypes[poolType]) {
+            throw new Error('Unsupported pool type');
```
brunoguerios, line 16: I believe that logging `poolType` here could be helpful:
```suggestion
            throw new Error('Unsupported pool type: ${poolType}');
```

src/entities/createPool/types.ts
```diff
@@ -0,0 +1,33 @@
+import { Address, Hex } from 'viem';
+
+export interface CreatePoolBase {
+    buildCall(input: CreatePoolInput): CreatePoolBuildCallOutput;
+}
+
+export type CreatePoolInput = WeightedCreatedPoolInput;
+
+export type WeightedCreatedPoolInput = {
```
brunoguerios, line 9: I see we have a few different naming standards on the code, for example `WeightedCreatedPoolInput` and `CreateWeightedPoolArgs`.

When refactoring Add/Remove Liquidity interface, I found it better to keep the same prefix everywhere, so for example:
- `WeightedJoinInput` -> `AddLiquidityWeightedInput`
- `NestedJoin` -> `AddLiquidityNested`

What do you think of following the same pattern of using `CreatePool...`?
- `CreatePoolBase`
- `CreatePoolInput`
- `CreatePoolWeightedInput`
- `CreatePoolWeightedArgs`

src/entities/createPool/utils/validateInputs.ts
```diff
@@ -0,0 +1,35 @@
+import { CreatePoolInput } from '../types';
+
+export class ValidateInputs {
```
brunoguerios, line 3: Any reason to create a ValidateInputs class? We've been using a function on add/remove liquidity and I think it would be nice to keep the same standard between them. What do you think?

src/entities/createPool/weighted/createPoolWeighted.ts
```diff
@@ -0,0 +1,66 @@
...
+            args,
+        });
+        return { call: encodedCall };
+    }
+
+    private parseCreateFunctionArgs(
+        input: WeightedCreatedPoolInput,
+    ): CreateWeightedPoolArgs {
+        const sortedTokenParams = input.tokens.map((tokenAddress) => {
+            let weight: number = 1 / input.tokens.length;
```
brunoguerios, line 26: What happens for pools with 3 tokens? Sum won't be exactly 1, right? Is that a problem?

src/entities/createPool/types.ts
```diff
@@ -0,0 +1,33 @@
...
+    name?: string;
+    symbol: string;
+    tokens: Address[];
+    weights?: { tokenAddress: Address; weight: number }[];
```
brunoguerios, line 13: Out of curiosity, don't you loose some precision if you set `weight` as `number` since it needs to have 18 decimals?

lgahdl: JS float number accepts up to 16 decimals, so users would lose the 17th and 18th, but I don't think this can be a problem, it's a very edge case, but a solution that could solve this precision problem and still deliver the benefits of the natural numbers is to use weight values like w1=80, w2=20, instead of w1=0.8, w2=0.2, for a 80/20 pool, for example

brunoguerios: Another question - is it ok to make `weights` optional when creating a Weighted pool? I think this should be required. How was that being handled in the previous SDK?

src/entities/createPool/weighted/createPoolWeighted.ts
brunoguerios: Connecting to the other comment, if you make `weights` required, than you won't have to worry about creating a test case for this.

src/entities/createPool/types.ts
lgahdl: 1 - In the previous SDK was mandatory, I just didn't make mandatory to reduce the complexity of the pool creation in the equally weights cases, you found confusing for the user to create a pool without passing weights? I made this way, with equally distribution of weights as default, because we have a lot of 50/50 pools.

About the weight being a string, you're talking about a string with the solidity format, like "500....00" or just 0.5?

Obs.: Now that you pointed it out, I think that the rateProvider being optional (zeroAddress by default) may not be a good idea as well, I don't know if it's common to use zeroAddress as the rateProvider, or if this is just for tests, what do you think?

src/entities/createPool/types.ts
brunoguerios: This is still not following the pattern - it should be `CreatePoolWeigthedInput` and `CreatePoolWeightedArgs`.
This will make more sense when you start implementing Composable Stable pool create and have `CreatePoolComposableStableInput` and `CreatePoolComposableStableArgs`.
It's easier to follow when the prefixes are the same.

src/entities/createPool/types.ts
```diff
@@ -0,0 +1,33 @@
...
+    swapFee: string;
+    poolOwnerAddress: Address;
+    salt?: Hex;
+};
+
+export type CreatePoolBuildCallOutput = {
+    call: Hex;
+};
+
+export type CreateWeightedPoolArgs = [
```
brunoguerios, line 24: This is still not following the pattern - it should be `CreatePoolWeigthedInput` and `CreatePoolWeightedArgs`.
This will make more sense when you start implementing Composable Stable pool create and have `CreatePoolComposableStableInput` and `CreatePoolComposableStableArgs`.
It's easier to follow when the prefixes are the same.

brunoguerios (changes requested):
I left some comments on how I think the interface could be improved. Let me know what you think.

src/lib/constants/deployment-addresses.ts
```diff
@@ -0,0 +1,7 @@
+import { ChainId } from '../../utils';
+
+export const DEPLOYMENT_ADDRESSES = {
+    [ChainId.MAINNET]: {
+        weightedFactoryV4: '0x897888115ada5773e02aa29f775430bfb5f34c51',
```
brunoguerios, line 5: This is duplicated from the constants file. Can we remove it?

lgahdl: Yeah, of course

brunoguerios (approved):
LGTM!

PR #144, opened 2023-11-24
Price Impact - ABA - Add Liquidity Single Token

brunoguerios:


src/entities/priceImpactAmount.ts
```diff
@@ -0,0 +1,35 @@
+import { formatEther, parseEther } from 'viem';
+import { BigintIsh } from './tokenAmount';
+
+export class PriceImpactAmount {
+    public amount: bigint;
+    public decimal: number;
+    public percentage: number;
```
lgahdl, line 7: Maybe would be useful to have comments about what these "percentage" and "bps" parameters stands for?

PR #145, opened 2023-11-27
Price Impact - ABA - Add Liquidity Unbalanced

brunoguerios:


src/entities/priceImpact/index.ts
```diff
@@ -45,4 +54,134 @@ export class PriceImpact {
...
+            bptIn: bptOut.toInputAmount(),
+            kind: RemoveLiquidityKind.Proportional,
+        };
+        const { amountsOut } = await removeLiquidity.query(
+            removeLiquidityInput,
+            poolState,
+        );
+
+        // diffs between unbalanced and proportional amounts
+        const diffs = amountsOut.map((a, i) => a.amount - amountsIn[i].amount);
```
lgahdl, line 86: In some smart contracts, as the Vault, and in the b-sdk test helpers is used the word "deltas" instead of "diffs", maybe it would be better to use "deltas" to keep the pattern

src/entities/priceImpact/index.ts
```diff
@@ -45,4 +54,134 @@ export class PriceImpact {
...
+                diffBPTs[givenIndex] = 0n;
+                diffs[resultIndex] = diffs[resultIndex] + resultAmount.amount;
+                diffBPTs[resultIndex] = await queryBPTForDiffAtIndex(
+                    resultIndex,
+                );
+            }
+            return minNegativeDiffIndex;
+        }
+
+        async function queryBPTForDiffAtIndex(i: number): Promise<bigint> {
```
lgahdl, line 171: Maybe would worth to have a comment for this function, the name is not very explicative, but I understood that this function queries the bptOut amount for an amount delta (amountOut - amountIn), given the token index, right?
Obs.: What do you think of the suggested name: "queryBptOutForSingleTokenDiff(tokenIndex: number)"?

src/entities/priceImpact/index.ts
```diff
@@ -45,4 +54,134 @@ export class PriceImpact {
...
+                    (diffBPT) =>
+                        diffBPT === min(diffBPTs.filter((a) => a > 0n)),
+                );
+                minNegativeDiffIndex = diffBPTs.findIndex(
+                    (diffBPT) =>
+                        diffBPT === max(diffBPTs.filter((a) => a < 0n)),
+                );
+
+                let kind: SwapKind;
+                let givenIndex: number;
```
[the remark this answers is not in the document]
brunoguerios: I agree that adding `token` to the name is helpful 👍 
But I'm not so sure about `received`, because I think it may be confused with `tokenOut`, which is not always the case. For `GivenOut` swaps, `received` would be related to `tokenIn`, which is a bit odd.
I decided to go with `result` because that's what is returned from the `query` action [here](https://github.com/balancer/b-sdk/blob/d0a871cd59855f156f249dab218b9ed175e39cc2/src/entities/utils/doQuerySwap.ts#L46)
With that in mind, I think we should go with `givenTokenIndex` and `resultTokenIndex`.
Let me know if that makes sense.

src/entities/priceImpact/index.ts
brunoguerios: Yeah, I agree this function name could be improved. We should likely make it more explicit that is an `addLiquidity` query as well.
Let me know what you think of this suggestion:
```suggestion
        async function queryAddLiquidityForTokenDelta(tokenIndex: number): Promise<bigint> {
```

PR #150, opened 2023-12-04
Add Liquidity Init - Weighted Pool

lgahdl:
- new class AddLiquidityInit;
- Encapsulating validators in a class InputValidator;
- moving encodeUserData functions to pool type encoders;
- addLiquidityInitWeighted tests;


src/data/providers/addLiquidityInitPoolDataProvider.ts
```diff
@@ -0,0 +1,75 @@
...
+        amounts: InputAmount[],
+        input: CreatePoolInput,
+    ): Promise<PoolStateInput> {
+        const poolContract = getContract({
+            abi: this.simplePoolAbi,
+            address: poolAddress,
+            publicClient: this.client,
+        });
+
+        const poolId = (await poolContract.read.getPoolId()) as Hex;
```
brunoguerios, line 55: What happens if an invalid address is provided? Maybe it's safer to throw in case it doesn't find a poolId?

src/data/providers/addLiquidityInitPoolDataProvider.ts
```diff
@@ -0,0 +1,75 @@
...
+        return {
+            id: poolId,
+            address: poolAddress,
+            type: poolType.toUpperCase(),
+            tokens: amounts
+                .sort((a, b) => {
+                    const diff = BigInt(a.address) - BigInt(b.address);
+                    return diff > 0 ? 1 : diff < 0 ? -1 : 0;
```
brunoguerios, line 63: You should be able to simplify with:
```suggestion
                    return BigInt(a.address) - BigInt(b.address);
```

brunoguerios (changes requested):
I think we could go a bit further on the idea of separating `init` from `add liquidity` and create a separate entity called `InitPool`, with the same structure as `CreatePool`. The way it's currently implemented, is still feels part of the add liquidity flow.
I'm not completely sure on the impact of this, but I'd like to know what do you think of this idea. Do you think it's worth evaluating that change?
Ps: I added a few small comments on the PR that would be nice to update as well.

lgahdl:
> I think we could go a bit further on the idea of separating `init` from `add liquidity` and create a separate entity called `InitPool`, with the same structure as `CreatePool`. The way it's currently implemented, is still feels part of the add liquidity flow. I'm not completely sure on the impact of this, but I'd like to know what do you think of this idea. Do you think it's worth evaluating that change? Ps: I added a few small comments on the PR that would be nice to update as well.

My first version of the Add Liquidity Init was a different class, but I noticed that was possible to reuse `parseAddLiquidityArgs` and `encodeUserData`, so I moved the functionality to AddLiquidity, but we can also do the other way around, that is bring the 2 reusable functions to outside the AddLiquidity model, no problem, we just cannot have duplicated code.

src/data/providers/addLiquidityInitPoolDataProvider.ts
lgahdl: Doesn't work because bigint is not accepted as a result, but bigint comparisons with numbers using < and > operators is possible

src/data/providers/addLiquidityInitPoolDataProvider.ts
```diff
@@ -0,0 +1,81 @@
...
+                        const diff = BigInt(a.address) - BigInt(b.address);
+                        return diff > 0 ? 1 : diff < 0 ? -1 : 0;
+                    })
+                    .map(({ address, decimals }, index) => ({
+                        address,
+                        decimals,
+                        index,
+                        weight: input.tokens.find(
+                            (t) => t.tokenAddress === address,
+                        )?.weight,
```
brunoguerios, line 71: Are you validating anywhere that amounts are being provided with the same address as CreatePoolInput? In other words, can this weight end up being undefined?
Another thing to keep in mind is that this check is case sensitive. Is that a problem?

src/entities/encoders/composableStable.ts
```diff
@@ -22,6 +25,78 @@ export class ComposableStableEncoder {
...
 
+    /**
+     * Encodes the User Data for adding liquidity to a ComposableStablePool
+     * @param kind Kind of the Add Liquidity operation: Init, Unbalanced, SingleToken, Proportional
+     * @param amounts Amounts of tokens to be added to the pool
+     * @returns
+     */
+    static encodeAddLiquidityUserData(
+        kind: AddLiquidityKind,
+        amounts: AddLiquidityAmounts & { maxAmountsInNoBpt: bigint[] },
```
brunoguerios, line 36: We've been using `withBpt` and `withoutBpt` - can you please update to keep the standard.

src/entities/inputValidator/inputValidator.ts
```diff
@@ -0,0 +1,49 @@
+import { AddLiquidityInput } from '../addLiquidity';
+import { CreatePoolInput } from '../createPool/types';
+import { PoolStateInput } from '../types';
+import { InputValidatorComposableStable } from './composableStable/inputValidatorComposableStable';
+import { InputValidatorGyro } from './gyro/inputValidatorGyro';
+import { InputValidatorBase } from './types';
+import { InputValidatorWeighted } from './weighted/inputValidatorWeighted';
+
+export class InputValidator {
+    validators: Record<string, InputValidatorBase> = {};
```
brunoguerios, line 10: I know this replicates how other classes are implementing this, but I think we could benefit from using a PoolType type instead of string. This can be done by altering the api-client to map poolType from `string` to `PoolType`. I'll create a task to evaluate this refactor in the future. As reference, Add/Remove Nested Liquidity PR already has this change in place.

src/types.ts
```diff
@@ -68,3 +68,9 @@ export type InputAmount = {
     decimals: number;
     rawAmount: bigint;
 };
+
+export type InitInputAmount = InputAmount | WeightedInitInputAmount;
+
+export type WeightedInitInputAmount = InputAmount & {
```
brunoguerios, line 74: I'd follow the reversed pattern we've been using:
- `InitInputAmount` -> `InputAmountInit`
- `WeightedInitInputAmount` -> `InputAmountWeighted`

brunoguerios (approved):
Awesome refactor! 👏 👏 👏 
I added some small comments, but none of them are blockers.
I do want to discuss one big thing though. Since we're trying to present init as a separate flow, I think we should move away from the AddLiquidity name - even though it uses the same encoders, validators, etc..
I believe that simply renaming it to `InitPool` instead of `AddLiquidityInit` would already be enough to achieve that.
What do you think?

src/data/providers/addLiquidityInitPoolDataProvider.ts
lgahdl: Actually the weight shouldn't be in there, I'm using the weight from the initInputAmount in the buildCall, not from the poolState, already removed

src/entities/inputValidator/inputValidator.ts
lgahdl: Yeah, I used it as string because I didn't change the PoolType class to UpperCase, api sends the type as WEIGHTED, PHANTOM_STABLE, GYRO2... etc, I can create an adapter to bring the returned pool types from API as an enum PoolType value, what do you think?

brunoguerios: I'd say it's ok to leave as is for now - I just created #157 so we don't forget to come back to this in the future.

src/entities/initPool/weighted/addLiquidityInitWeighted.ts
```diff
@@ -0,0 +1,84 @@
...
+            minimumBpt,
+            maxAmountsIn: getAmounts(
+                sortTokensByAddress(poolTokens),
+                input.amountsIn,
+            ),
+            tokenInIndex: undefined,
+        };
+    }
+
+    private calculateBptOut(amounts: InputAmountInitWeighted[]): bigint {
```
brunoguerios, line 72: If I understood correctly this means that we'll need to know each pool math to be able to implement InitPool, right?
We're trying to avoid the need to know pool's math everywhere within b-sdk, so it's simpler to implement.
Are we able to avoid it somehow.?

src/entities/initPool/weighted/addLiquidityInitWeighted.ts
```diff
@@ -0,0 +1,84 @@
...
+        const tokensQtd = amounts.length;
```
brunoguerios, line 73: What does `tokensQtd` mean? Maybe `tokensLength` is a bit more intuitive?

src/entities/initPool/weighted/addLiquidityInitWeighted.ts
```diff
@@ -0,0 +1,84 @@
...
+        const poolInvariant = amounts.reduce((acc, curr) => {
+            return (
+                acc *
+                parseFloat(formatUnits(curr.rawAmount, curr.decimals)) **
+                    parseFloat(formatUnits(BigInt(curr.weight), 18))
```
brunoguerios, line 78: I'm not super comfortable using float point math for fixed point values like this. You will likely lose precision. How was this handled in the previous SDK?

src/entities/inputValidator/inputValidator.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    constructor() {
+        this.validators = {
+            WEIGHTED: new InputValidatorWeighted(),
+            GYRO2: new InputValidatorGyro(),
+            GYRO3: new InputValidatorGyro(),
+            GYROE: new InputValidatorGyro(),
+            PHANTOM_STABLE: new InputValidatorComposableStable(),
+        };
+    }
```
brunoguerios, line 20: Since we're replicating the same structure everywhere, what do you think of adding the ability to add customValidators as well?

test/lib/utils/initPoolHelper.ts
```diff
@@ -0,0 +1,82 @@
...
+        initPoolBuildOutput.call,
+        initPoolBuildOutput.value,
+    );
+    return {
+        initPoolBuildOutput,
+        txOutput,
+    };
+}
+
+export function assertAddLiquidityInit(
```
brunoguerios, line 56: There are some `addLiquidityInit` left in the code.

src/entities/inputValidator/inputValidator.ts
lgahdl: It wouldn't make much sense I think, since the InputValidator is not exposed to the user

src/entities/initPool/weighted/addLiquidityInitWeighted.ts
lgahdl: There's no queryJoin for init, because the invariant of the pool is zero, so I'd say there's no way to get the bptOut other than calculate it 🫤 but we can remove the bptOut from the return parameters

src/entities/inputValidator/inputValidator.ts
brunoguerios: But the idea is that devs should be able to extend functionality such as addLiquidity and removeLiquidity - this means that they would want to extend their respective validator, right?
Anyway, that's basically a helper, so I'd say it's ok to leave as is and reassess in the future 👍

src/entities/initPool/weighted/addLiquidityInitWeighted.ts
brunoguerios: Yes, I agree we should remove 👍

src/entities/initPool/weighted/addLiquidityInitWeighted.ts
```diff
@@ -0,0 +1,84 @@
...
+import { TokenAmount } from '../../tokenAmount';
+import { sortTokensByAddress } from '../../../utils/tokens';
+import { vaultAbi } from '../../../abi';
+import { getAmounts, parseAddLiquidityArgs } from '../../utils';
+import { Token } from '../../token';
+import { WeightedEncoder } from '../../encoders';
+import { AddLiquidityBuildOutput } from '../../addLiquidity/types';
+import { InputAmountInitWeighted } from '../../../types';
+
+export class AddLiquidityInitWeighted implements InitPoolBase {
```
brunoguerios: `InitPoolWeighted`

src/entities/initPool/weighted/addLiquidityInitWeighted.ts
```diff
@@ -0,0 +1,84 @@
...
+    buildCall(
+        input: InitPoolInput,
+        poolState: PoolState,
+    ): AddLiquidityBuildOutput {
```
brunoguerios: `InitBuildOutput`

src/entities/addLiquidity/types.ts
```diff
@@ -40,7 +36,7 @@ export type AddLiquidityProportionalInput = AddLiquidityBaseInput & {
 };
 
 export type AddLiquidityInput =
-    | AddLiquidityInitInput
+    | InitPoolInput
```
brunoguerios, line 39: You should remove this option from here.

src/entities/addLiquidity/types.ts
```diff
@@ -103,6 +93,11 @@ export type AddLiquidityBuildOutput = {
     maxAmountsIn: TokenAmount[];
 };
 
+export type InitPoolBuildOutput = Omit<
+    AddLiquidityBuildOutput,
+    'minBptOut' | 'maxAmountsIn'
+>;
```
brunoguerios, line 99: I think this makes more sense in the init pool types file - what do you think?

brunoguerios (changes requested):
Still some minor stuff left. Hopefully it's the last one 🙏

src/entities/initPool/weighted/addLiquidityInitWeighted.ts
brunoguerios: ignore - I was looking to an outdated code ✅

src/entities/encoders/composableStable.ts
```diff
@@ -22,6 +25,78 @@ export class ComposableStableEncoder {
...
+     */
+    static encodeAddLiquidityUserData(
+        kind: AddLiquidityKind,
+        amounts: AddLiquidityAmounts & { maxAmountsInWithoutBpt: bigint[] },
+    ): Address {
+        switch (kind) {
+            case AddLiquidityKind.Init:
+                return ComposableStableEncoder.addLiquidityInit(
+                    amounts.maxAmountsInWithoutBpt,
+                );
```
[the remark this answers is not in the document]
lgahdl: I'm doing this on another PR along with the composable stable create and initPool

brunoguerios (approved):

PR #159, opened 2023-12-11
Composable Stable - Create Pool

lgahdl:
Adding CreatePool class for Composable Stable pools; Adding tests;
Adding constants and abi;
Renaming "createWeightedPool.integration.test.ts" to "createPoolWeighted.integration.test.ts"; Adapting createPoolHelper to new poolTypes;


src/entities/createPool/types.ts
```diff
@@ -19,6 +21,21 @@ export type CreatePoolWeightedInput = {
     salt?: Hex;
 };
 
+export type CreatePoolComposableStableInput = {
```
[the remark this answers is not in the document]
brunoguerios: We'll handle only the most recent pool version/factory within the SDK. 
But your question does highlight that this info is not yet clear enough.
Maybe we should mention this on readme/comments/examples/docs?

src/entities/inputValidator/utils/validateTokens.ts
```diff
@@ -68,3 +68,18 @@ export const validatePoolHasBpt = (poolState: PoolStateInput) => {
...
+export const validateCreatePoolTokens = (
+    tokens: { tokenAddress: string }[],
+) => {
+    const tokenAddresses = tokens.map((t) => t.tokenAddress);
+    if (tokenAddresses.length !== new Set(tokenAddresses).size) {
+        throw new Error('Duplicate token addresses');
+    }
+    if (tokens.length > 4) {
+        throw new Error('Maximum of 4 tokens allowed');
+    }
```
brunoguerios, line 81: Can you please check how clear is the error returned from the SC to a request with > tokens than allowed?
If we consider it's good enough, we could simply let it fail at the SC level, right?
If we think the error message can be improved, then it would be nice to replicate the limit the SC defines (5 tokens on CSP and 8 tokens on weighted pools, right?)

lgahdl: Done

src/entities/createPool/types.ts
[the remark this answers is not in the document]
lgahdl: Added 2 examples and mentioned that only composable stable v5 and weighted v4 pools are available

src/entities/createPool/createPool.ts
```diff
@@ -11,6 +12,7 @@ export class CreatePool {
         this.inputValidator = new InputValidator();
         this.createPoolTypes = {
             WEIGHTED: new CreatePoolWeighted(),
+            PHANTOM_STABLE: new CreatePoolComposableStable(),
```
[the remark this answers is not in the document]
lgahdl: I'll take this task to map the pool types as the next one, I totally agree that use PHANTOM_STABLE is very confusing in the code

src/abi/composableStableFactoryV5.ts
```diff
@@ -0,0 +1,186 @@
...
+        type: 'function',
+    },
+    {
+        inputs: [],
+        name: 'version',
+        outputs: [{ internalType: 'string', name: '', type: 'string' }],
+        stateMutability: 'view',
+        type: 'function',
+    },
+];
```
brunoguerios, line 186:
```suggestion
] as const;
```

src/utils/constants.ts
```diff
@@ -173,6 +173,16 @@ export const WEIGHTED_POOL_FACTORY: Record<number, Address> = {
     [ChainId.ZKEVM]: '0x03f3fb107e74f2eac9358862e91ad3c692712054',
 };
 
+export const COMPOSABLE_STABLE_POOL_FACTORY: Record<number, Address> = {
+    [ChainId.ARBITRUM_ONE]: '0xa8920455934da4d853faac1f94fe7bef72943ef1',
```
brunoguerios, line 177: Can you please add a comment with a link to the source for these addresses?

src/abi/composableStableFactoryV5.ts
brunoguerios: Please to the same for Weighted pool factory abi

brunoguerios (commented):
LGTM!
As soon as you add a changeset and address these minor comments I added, we should be ready to merge/release.

PR #160, opened 2023-12-11
refactor: load shared abi from json files

mkflow27:
This pr replaces the human readable abi with JSON abis loaded to create a shared abi.


src/data/onChainPoolDataViaReadContract.ts
```diff
@@ -18,24 +30,22 @@ type Result =
...
+    ...GyroEV2Abi,
+    ...LiquidityBootstrappingPoolAbi,
+    ...ManagedPoolAbi,
+    ...MetaStablePoolAbi,
+    ...PhantomStablePoolAbi,
+    ...WeightedPoolAbi,
+    ...VaultAbi
+]
+
+const uniqueAbiElements = new Map(abis.map(abi => [JSON.stringify(abi), abi]));
```
brunoguerios, line 47: I'm sorry, but I just realized the existing abis are not `.json`, but `.ts` files exported as `const`.
I think we should benefit from keeping the same standard. 
Can you please update your implementation to that instead?

src/data/onChainPoolDataViaReadContract.ts
```diff
@@ -18,24 +30,22 @@ type Result =
...
+    ...ConfigurableRightsPoolAbi,
+    ...ConvergentCurvePoolAbi,
+    ...FxPoolAbi,
+    ...GyroEV2Abi,
+    ...LiquidityBootstrappingPoolAbi,
+    ...ManagedPoolAbi,
+    ...MetaStablePoolAbi,
+    ...PhantomStablePoolAbi,
+    ...WeightedPoolAbi,
+    ...VaultAbi
```
brunoguerios, line 44: You should be able to reuse the existing vault abi if you move to .ts instead of .json abis.

src/data/onChainPoolDataViaReadContract.ts
```diff
@@ -1,8 +1,20 @@
...
+import ConfigurableRightsPoolAbi from '../abi/abis/ConfigurableRightsPoolAbi.json';
+import ConvergentCurvePoolAbi from '../abi/abis/ConvergentCurvePoolAbi.json';
+import FxPoolAbi from '../abi/abis/FxPoolAbi.json';
+import GyroEV2Abi from '../abi/abis/GyroEV2Abi.json';
+import LiquidityBootstrappingPoolAbi from '../abi/abis/LiquidityBootstrappingPoolAbi.json';
+import ManagedPoolAbi from '../abi/abis/ManagedPoolAbi.json';
+import MetaStablePoolAbi from '../abi/abis/MetaStablePoolAbi.json';
+import PhantomStablePoolAbi from '../abi/abis/PhantomStablePoolAbi.json';
+import VaultAbi from '../abi/abis/VaultAbi.json';
+import WeightedPoolAbi from '../abi/abis/WeightedPoolAbi.json';
```
brunoguerios, line 16: It would be nice if you could move these abis to the parent folder and export them on index.ts, so it's easier to import them elsewhere.

src/data/onChainPoolDataViaReadContract.ts
```diff
@@ -1,8 +1,20 @@
-import { PublicClient, parseAbi, Address, Abi, Hex } from 'viem';
+import { PublicClient, Address, Abi, Hex } from 'viem';
 import { getPoolAddress } from '../utils';
 import { OnChainPoolData } from './enrichers/onChainPoolDataEnricher';
 import { SwapOptions } from '../types';
 
+import ComposableStablePoolV5Abi from '../abi/abis/ComposableStablePoolV5Abi.json';
+import ConfigurableRightsPoolAbi from '../abi/abis/ConfigurableRightsPoolAbi.json';
+import ConvergentCurvePoolAbi from '../abi/abis/ConvergentCurvePoolAbi.json';
```
brunoguerios, line 8: I'm not familiar with these abis - can you give more details on where you did you get them?

brunoguerios (changes requested):
Hey! Sorry for the wrong task specs. 😬  
I hope you don't mind updating the implementation to match the existing pattern.
Let me know if you have any questions!

src/data/onChainPoolDataViaReadContract.ts
mkflow27: included

src/data/onChainPoolDataViaReadContract.ts
mkflow27: I have moved the abis to be exported from .ts files.

src/data/onChainPoolDataViaReadContract.ts
mkflow27: yes. removed the old one.

src/abi/composableStablePoolV5.ts
```diff
@@ -0,0 +1,840 @@
+export const composabableStablePoolV5Abi = [
+    [
```
brunoguerios, line 2: There extra brackets here that should be removed.

src/data/onChainPoolDataViaReadContract.ts
```diff
@@ -18,25 +37,6 @@ type Result =
...
-    'function getWrappedTokenRate() view returns (uint256)',
-    'function getAmplificationParameter() view returns (uint256 value, bool isUpdating, uint256 precision)',
-    'function getPausedState() view returns (bool)',
-    'function inRecoveryMode() view returns (bool)',
-    'function getRate() view returns (uint256)',
-    'function getScalingFactors() view returns (uint256[] memory)',
-]);
-
 // Extract the functionName property values into a union type
 type FunctionNameUnion = typeof abi[number]['name'];
```
brunoguerios, line 41: I checked out your branch and it seems that this type is not being computed properly.
Can you fix it?
If it starts taking too long, I'd say it's ok to leave as is, since all tests and lint checks are passing.

src/abi/GyroEV2.ts
```diff
@@ -0,0 +1,1970 @@
+export const gyroEV2Abi = [
+        {
+            inputs: [
```
brunoguerios, line 3: Indentation here is not properly set. Maybe it's a prettier issue? Can you double check this on the abi files you added?
I know it seems a nuisance, but if I simply open and close the file on my end, I add changes to the whole file 😅

mkflow27: Thanks for highlithing. I have adapted my prettier settings.

src/abi/composableStablePoolV5.ts
mkflow27: done

src/abi/GyroEV2.ts
brunoguerios: I'm not really sure what happened. Abi files were formatted correctly, but `onChainPoolDataViaReadContract.ts` was not. I updated the file over here, but it would be nice to double check your prettier next time you make a change 😉.

brunoguerios (approved):

PR #167, opened 2023-12-12
152 composable stable pool init pool

lgahdl:
Creating Init Pool Functionality for Composable Stable pools;
Tests for Composable Pool Initialization;


src/entities/addLiquidity/composable-stable/addLiquidityComposableStable.ts
```diff
@@ -206,7 +205,9 @@ export class AddLiquidityComposableStable implements AddLiquidityBase {
                 break;
             }
             default:
-                throw Error('Unsupported Add Liquidity Kind');
+                throw Error(
+                    'Unsupported Add Liquidity Kind, for Init use InitPool instead of AddLiquidity',
+                );
```
brunoguerios, line 210: I think it's a bit odd to add the Init specific error message to the default case.
How about we keep the Init case and move the specific Init error message there?

src/entities/initPool/composableStable/initPoolComposableStable.ts
```diff
@@ -0,0 +1,73 @@
...
+        const sortedTokens = sortTokensByAddress(poolState.tokens);
+        const bptIndex = sortedTokens.findIndex(
+            (t) => t.address === poolState.address,
+        );
+        const maxAmountsIn = getAmounts(sortedTokens, [
+            ...input.amountsIn.slice(0, bptIndex),
+            {
+                address: poolState.address,
+                decimals: 18,
+                rawAmount: BigInt.asUintN(256, BigInt(-1)),
```
brunoguerios, line 55: Is that `MAX_UINT256`? I think it's easier to understand if you use the constant instead.

src/data/providers/initPoolDataProvider.ts
```diff
@@ -49,22 +50,39 @@ export class InitPoolDataProvider {
...
+            }));
+
+        const poolTokensWithBpt = sortTokensByAddress([
+            {
+                address: poolAddress.toLowerCase() as Address,
+                decimals: 18,
+                index: 0,
+            },
+            ...poolTokens,
+        ]);
```
brunoguerios, line 72: Since BPT is not necessarily at index 0, you should build/sort the tokens array only after adding the BPT, right?

brunoguerios (commented):
Overall LGTM!
I added some minor comments that would be nice to address before approving/merging.
Please add the changeset as well, so we can get it ready for merging afterwards

src/data/providers/initPoolDataProvider.ts
lgahdl: That's what's happening, it's sorting the array with the BPT token inside

src/data/providers/initPoolDataProvider.ts
```diff
@@ -49,22 +50,35 @@ export class InitPoolDataProvider {
...
+                address: address.toLowerCase() as Address,
+                decimals,
+                index: index + 1,
+            }));
+
+        const poolTokensWithBpt = sortTokensByAddress([
+            {
+                address: poolAddress.toLowerCase() as Address,
+                decimals: 18,
+                index: 0,
```
brunoguerios, line 65: Is this truly always zero? I was under the impression it would be index of its position on the sorted array.

src/entities/addLiquidity/weighted/addLiquidityWeighted.ts
```diff
@@ -143,7 +146,9 @@ export class AddLiquidityWeighted implements AddLiquidityBase {
                 };
             }
             default:
-                throw Error('Unsupported Add Liquidity Kind');
+                throw Error(
+                    'Unsupported Add Liquidity Kind, for Init use InitPool instead of AddLiquidity',
+                );
```
brunoguerios, line 151: For the default case, I'd go with a generic error message, without mentioning Init:
```suggestion
                throw Error(
                    'Unsupported Add Liquidity Kind: ${input.kind}',
                );
```

src/entities/addLiquidity/weighted/addLiquidityWeighted.ts
```diff
@@ -143,7 +146,9 @@ export class AddLiquidityWeighted implements AddLiquidityBase {
                 };
             }
             default:
-                throw Error('Unsupported Add Liquidity Kind');
+                throw Error(
+                    'Unsupported Add Liquidity Kind: ${input.kind}',
```
brunoguerios, line 150:
```suggestion
                    `Unsupported Add Liquidity Kind: ${input.kind}`,
```

brunoguerios (approved):

PR #170, opened 2023-12-14
chore: expose priceImpact amount from entities index

agualis:
We need `PriceImpactAmount` in the consumers to get the return type of `PriceImpact.addLiquidityUnbalanced` but I noticed that it was not exposed.


brunoguerios (approved):
Thanks for the patch 😉

PR #180, opened 2023-12-20
V3 Scaffold (alternative)

brunoguerios:


src/data/providers/initPoolDataProvider.ts
```diff
@@ -80,6 +80,7 @@ export class InitPoolDataProvider {
                 address: poolAddress.toLowerCase() as Address,
                 type: poolType,
                 tokens: tokensPerPoolType[poolType],
+                balancerVersion: 2, // TODO V3: instantiate a different provider for V3? Or add a config/input to this one? Will the interface be the same?
```
danielmkm, line 83: Pools probably won't have a `getPoolId` function, so this would need to be different

danielmkm (approved):
Really like the approach here, also like the single entry point for V2 vs V3 👏

src/data/providers/initPoolDataProvider.ts
brunoguerios: Ok, I'll break this into 2 providers and leave the v3 implementation blank for now 👍

brunoguerios: In fact I'll leave that to be done in a separate PR because InitPoolDataProvider will require a bigger refactor/fix to handle all supported pool types

PR #181, opened 2023-12-20
Refactor to prepare for v3 support

brunoguerios:
I identified a few refactors that can help a lot on the v2/v3 support.
They are pretty straightforward, but they do add a lot of changes that would pollute the v3 scaffold PR, so I decided to push them separately.
Here are the main changes:
- Merge PoolStateInput and PoolState into PoolState
    - This enables AddLiquidity to implement AddLiquidityBase, making it easier to organize between v2 and v3 since all layers implement the same interface
    - replicate the same pattern on all entities (removeLiquidity, initPool, etc..)
- Enable absolute paths
    - more robust and easy to organize imports


danielmkm (approved):
Looks good Bruno. I like that `PoolState` is now using the minimal token now, it's easier to construct 🙌

PR #192, opened 2024-01-10
Replacing BalancerHelpers by BalancerQueries contract;

lgahdl:
Removing BalancerHelpers contract and replacing its use by BalancerQueries, since BalancerQueries have all the functions that are being used from BalancerHelpers.


brunoguerios:
The tasks actually expects the opposite, which is to stop using BalancerQueries, as it's actually another layer on top of BalancerHelpers.
We were using it as a helper contract that performs many operations, but since we stopped using it in favor of MulticallV3, I believe it makes sense to rely only on BalancerHelpers directly.
queryBatchSwap can be done directly at the Vault level, so we should be ok without BalancerQueries.

brunoguerios (approved):

brunoguerios (approved):

PR #195, opened 2024-01-11
Fixing InitPoolDataProvider to work with any pool type;

lgahdl:


brunoguerios:
Looking closer at this provider, I feel like it's a bit fragile in the sense that we rely mostly on user input to build the pool state, without some validation/confirmation that the info is correct.
If you're already relying on onChain data to fetch the poolId, we could also:
- use `getPoolTokens` method from the Vault to fetch all token addresses (and their indexes)
- use ERC20 contract to fetch token decimals

With this approach the user doesn't even need to know specify if the pool contains BPT or not, because getPoolTokens returns everything you need.
What do you think?

src/utils/tokens.ts
```diff
@@ -1,3 +1,32 @@
...
+        },
+    ];
+
+    const tokenContract = getContract({
+        abi,
+        address: tokenAddress,
+        publicClient: client,
+    });
+    const decimals = await tokenContract.read.decimals();
+    console.log(decimals);
```
brunoguerios, line 30: Is this a debug log? I'd say we should remove if you won't need them anymore.

brunoguerios (approved):
It's way better like this 👏 
There is only a small comment there - feel free to merge after responding to it.

src/utils/tokens.ts
lgahdl, line 30:
```suggestion
```

lgahdl: Sorry, didn't notice that, thanks!

PR #204, opened 2024-01-16
Add Normalized Liquidity for all pool types

### LINKED ISSUE #203: Add Normalized Liquidity for all pool types
Pool types:

- [ ] Weighted
- [ ] Composable Stable
- [ ] Gyro Pools
- [ ] Fx Pools

brunoguerios:
This PR should serve as a place for discussion/iteration on normalized liquidity implementation.

Normalized liquidity is being used by the SOR to compare paths and find the best candidates (with deepest liquidity).
On the previous SOR we relied on the derivative spot price to calculate that, but the current SOR is aiming for a more simplified solution.
Key thing to keep in mind is that we need to find comparable values between different pool types, so we're able to properly select the best path/route for swaps.

Closes #203


brunoguerios:
Closing this PR as the implementation was moved to the API

PR #211, opened 2024-01-19
Reproduce issue with nested liquidity and WETH

agualis:
This PR has 2 things: 

1. It adds a new failing test:` trying to add nested liquidity with single token WETH` throws with `BAL#001` error
2. Little refactor to extract token addresses and cleanup nested tests (it could be done in other tests too)


brunoguerios:
@agualis - thanks for opening this issue. The problem is that when creating the relayer actions, I have to prune/skip the ones that have all amounts in equal zero.
In the test case you created, the join action at the 3POOL should be removed from the chain because only WETH was being provided as input directly to WETH_3POOL.
I also refactored a bit the addresses/constants file with something a bit more similar to what we had in the previous SDK, which is a bit more structured to support testing on multiple chains.
Let me know what you think and I'll replicate that to other tests through the code.

agualis:
@brunoguerios Looks good 👍

PR #219, opened 2024-01-25
Basic Add Liquidity for V3

### LINKED ISSUE #217: V3: Basic Add Liquidity

brunoguerios:
Closes #217


src/utils/constants.ts
```diff
@@ -58,12 +59,13 @@ export enum ChainId {
...
     ZKSYNC_TESTNET = 280,
+    FANTOM = 250,
     ZKSYNC = 324,
     ZKEVM = 1101,
     ARBITRUM_ONE = 42161,
     AVALANCHE = 43114,
     BASE_GOERLI = 84531,
-    FANTOM = 250,
+    SEPOLIA = 11155111,
 }
```
lgahdl, line 69: Maybe this could be sorted in alphabetical order instead of numerical?

lgahdl (approved):

src/utils/constants.ts
brunoguerios: I don't have a strong opinion on this, but I'd say it's ok as is.

src/entities/addLiquidity/addLiquidityV3/index.ts
```diff
@@ -0,0 +1,154 @@
...
+} from '../types';
+import { doAddLiquidityUnbalancedQuery } from './doAddLiquidityUnbalancedQuery';
+import { doAddLiquiditySingleTokenQuery } from './doAddLiquiditySingleTokenQuery';
+
+export class AddLiquidityV3 implements AddLiquidityBase {
+    async query(
+        input: AddLiquidityInput,
+        poolState: PoolState,
+    ): Promise<AddLiquidityBaseQueryOutput> {
+        const sortedTokens = getSortedTokens(poolState.tokens, input.chainId);
```
johngrantuk, line 32: Do we still need to sort tokens?

src/entities/addLiquidity/addLiquidityV3/index.ts
```diff
@@ -0,0 +1,154 @@
...
+                }
+                break;
+        }
+
+        let value = 0n;
+        if (input.wethIsEth) {
+            const wethInput = input.amountsIn.find(
+                (a) => a.token.address === NATIVE_ASSETS[input.chainId].wrapped,
+            );
+            value = wethInput?.amount ?? 0n;
```
johngrantuk, line 138: Do you think we should throw if wethInput doesn't exist?

src/entities/encoders/composableStable.ts
```diff
@@ -50,17 +51,16 @@ export class ComposableStableEncoder {
...
                 return ComposableStableEncoder.addLiquidityUnbalanced(
                     amounts.maxAmountsInWithoutBpt,
                     amounts.minimumBpt,
                 );
             case AddLiquidityKind.SingleToken: {
-                if (amounts.tokenInIndex === undefined) throw Error('No Index');
+                // just a sanity check as this is already checked in InputValidator
+                if (amounts.tokenInIndex === undefined) {
+                    throw addLiquiditySingleTokenShouldHaveTokenInIndexError;
+                }
```
johngrantuk, line 63: Do we still need it then? 😉

src/entities/addLiquidity/addLiquidityV3/index.ts
brunoguerios: Yeap. It would be odd if the flag `wethIsEth` is set to true and `wethInput` doesn't exist. I agree it's safer to throw and let the user know there's something wrong. 👍

src/entities/encoders/composableStable.ts
brunoguerios: There are a few ways to avoid it, but they are not necessarily better than the current solution 😅 
1. Add a bang operator - I wouldn't go that way, because in case we ever reach that error, it would fail without a proper message
2. Refactor the getAmountsCall to be specific for each RemoveLiquidityKind - which could increase the complexity given RemoveLiquidityCall is already a union type handling different pool types 😬

Do you have a preferred solution? Or even a 3rd option? 🙃

src/entities/addLiquidity/addLiquidityV3/index.ts
brunoguerios: done ✅

src/entities/encoders/composableStable.ts
johngrantuk: Ahh sorry, I meant if its already checked in InputValidator why do we have to check it again.

brunoguerios: Since this is non critical, I'll push in order to get things going. We can always refactor in a later PR. 🙃

johngrantuk: Makes sense, thanks! Seems good to just stick as is.

PR #220, opened 2024-01-25
Add limits calculation and data encoding in the swap

gmbronco:


test/entities/swap.test.ts
```diff
@@ -0,0 +1,117 @@
+import { SwapKind } from '@/types';
+import { Swap, PathWithAmount, Token, TokenAmount } from '@/entities';
+import { weightedPoolFactory } from '@/../test/factories/entities/weightedPool';
+
+describe('Swap', () => {
+    describe('limits', () => {
+        const tokens = [
+            new Token(1, '0x177127622c4a00f3d409b75571e12cb3c8973d3c', 18),
```
brunoguerios, line 8: Same here about using constants in a way that's easier to read which tokens are being used for testing.

debug/sor.ts
```diff
@@ -1,36 +1,141 @@
...
+const chainId = mainnet.id;
+
+const rpcUrl = {
+    [mainnet.id]: 'http://127.0.0.1:8545',
+    // [mainnet.id]: 'https://eth-mainnet.g.alchemy.com/v2/iOZ9hhr32Bq9txKoA8CZdcD7G2c7VARr',
+    [optimism.id]:
+        'https://opt-mainnet.g.alchemy.com/v2/iOZ9hhr32Bq9txKoA8CZdcD7G2c7VARr',
+    [gnosis.id]: 'https://rpc.gnosis.gateway.fm',
+    [fantom.id]: 'https://lb.nodies.app/v1/beefd211f635427099df316b5a737540',
+}[chainId];
```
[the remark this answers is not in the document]
gmbronco: good point, somehow i missed that debug scripts are checked in. removed the file.

src/entities/swap.ts
```diff
@@ -206,6 +207,144 @@ export class Swap {
...
     }
 
+    /**
+     * Takes a slippage acceptable by the user and returns the limits for a swap to be executed
+     *
+     * @param slippage percentage: 0.05 for 5%
+     * @param expectedAmount is the amount that the user expects to receive or send, can be obtained from swap.query()
+     * @returns
+     */
+    limits(slippage: number, expectedAmount: TokenAmount): bigint[] {
```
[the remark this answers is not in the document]
gmbronco: Am I missing something here? Slippage works only with 18 decimals and we'd need flexibility here to cover other coins.

src/entities/swap.ts
```diff
@@ -206,6 +207,144 @@ export class Swap {
...
+     * @param limits calculated from swap.limits()
+     * @param deadline unix timestamp
+     * @param sender address of the sender
+     * @param recipient defaults to sender
+     * @returns
+     */
+    private callData(
+        limits: bigint[],
+        deadline: bigint,
+        sender: `0x${string}`,
```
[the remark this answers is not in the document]
gmbronco: Not sure what do you mean by alias, sender type?

src/entities/swap.ts
```diff
@@ -206,6 +207,144 @@ export class Swap {
...
+        recipient = sender,
+    ): `0x${string}` {
```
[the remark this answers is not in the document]
gmbronco: good point, done.

src/entities/swap.ts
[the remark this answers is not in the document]
gmbronco: I see, decimals don't matter, because math sol is scaled to 18 dec as well, right?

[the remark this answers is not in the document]
brunoguerios: My suggestion was actually to expect Slippage as the input type, in the same way we're using TokenAmount. This way the user doesn't have to check units or convert whatever they have available. They can use any of variations to build a Slippage object and provide it.
Having it only internally doesn't seem to add that much value.

test/entities/swap.test.ts
```diff
@@ -0,0 +1,128 @@
...
+                        decimals: tokens[3].decimals,
+                        symbol: 'XXX',
+                        name: 'XXX Coin',
+                        index: 1,
+                    },
+                ],
+            }),
+        ];
+
+        test('amountIn to be maximally 10% more then expected', () => {
```
brunoguerios, line 99: Should be 1% based on the value you're testing, right?

test/entities/swap.test.ts
```diff
@@ -0,0 +1,128 @@
...
+                swapKind: SwapKind.GivenOut,
+            });
+
+            const slippage = 1;
+            const limits = swap.limits(slippage, amountOut);
+
+            expect(limits[0]).toEqual(1010000n);
+        });
+
+        test('amountOut to be maximally 10% less then expected', () => {
```
brunoguerios, line 112: Should be 1% based on the value you're testing, right?

src/entities/swap.ts
```diff
@@ -206,6 +207,145 @@ export class Swap {
...
+        sender: Address,
+        recipient = sender,
+    ): Hex {
+        let callData: Hex;
+
+        const funds = {
+            sender,
+            recipient,
+            fromInternalBalance: false,
+            toInternalBalance: false,
```
brunoguerios, line 322: Shouldn't we allow these to be set by the user?

test/entities/swap.test.ts
```diff
@@ -0,0 +1,128 @@
...
+            );
+            const swap = new Swap({
+                paths: [new PathWithAmount(tokens, pools, amountIn)],
+                swapKind: SwapKind.GivenIn,
+            });
+
+            const slippage = 1;
+            const limits = swap.limits(slippage, amountIn);
+
+            expect(limits[3]).toEqual(-990099009900990099n);
```
brunoguerios, line 125: Shouldn't it be -990000000000000000n? 🤔

src/entities/swap.ts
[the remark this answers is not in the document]
gmbronco: Sorry, i'm struggling with the internal types a bit. It mean that I'd need to import Slippage into the app as well, but I still need to know what units I'm dealing with for the initiation, so it's not really helping that much either.

test/entities/swap.test.ts
gmbronco: exactly, good catch!

src/entities/swap.ts
gmbronco: true, i'll set them to false by default. no one seems to be using these.

test/entities/swap.test.ts
gmbronco: I was surprised as well, but that's how the math in the Slippage object works. Probably needs a fix, wasn't sure.

debug/sor.ts
brunoguerios, line 1: Btw - I missed this, but this file is quite useful for debugging. Please revert your changes instead of deleting it 🙏

src/entities/slippage.ts
```diff
@@ -39,6 +39,6 @@ export class Slippage {
     }
 
     public removeFrom(amount: bigint): bigint {
-        return MathSol.divDownFixed(amount, this.amount + WAD);
+        return MathSol.mulDownFixed(amount, WAD - this.amount);
```
brunoguerios, line 42: I pushed a fix for this. Hope it's ok.

test/entities/swap.test.ts
```diff
@@ -0,0 +1,129 @@
...
+
+        test('amountIn to be maximally 1% more then expected', () => {
+            const amountOut = TokenAmount.fromRawAmount(tokens[3], 1000000n);
+            const swap = new Swap({
+                paths: [new PathWithAmount(tokens, pools, amountOut)],
+                swapKind: SwapKind.GivenOut,
+            });
+
+            const slippage = 1;
+            const limits = swap.limits(slippage, amountOut);
```
brunoguerios, line 108: I believe you should pass the `expectedAmountIn` as input to the limits calculation, right? 
One way of checking that is that the output should be in 18 decimals as the tokenIn is WETH for this test case.

test/entities/swap.test.ts
```diff
@@ -0,0 +1,129 @@
...
+                tokens[0],
+                1000000000000000000n,
+            );
+            const swap = new Swap({
+                paths: [new PathWithAmount(tokens, pools, amountIn)],
+                swapKind: SwapKind.GivenIn,
+            });
+
+            const slippage = 1;
+            const limits = swap.limits(slippage, amountIn);
```
brunoguerios, line 124: Same here.
I believe you should pass the `expectedAmountOut` as input to the limits calculation.
One way of checking that is that the output should be in 6 decimals as the tokenOut is USDT for this test case.

brunoguerios (changes requested):
Some minor comments, but tests should be fixed.

test/entities/swap.test.ts
lgahdl: I agree with Bruno about the amount variable name, if the swap kind is "givenOut", the variable name should be "amountIn".

About the test of the 18 decimals, I'm not so sure if this is helpful, because the result can be a fraction with less than 18 decimals

brunoguerios: It's not just a matter of variable name. We need to get `inputAmount` from swap object, apply slippage to it and check if that is what we expect it to be.

brunoguerios (approved):

PR #232, opened 2024-02-02
refactor: making amount smaller by slippage

gmbronco:


src/entities/slippage.ts
```diff
@@ -27,18 +27,30 @@ export class Slippage {
...
-    public removeFrom(amount: bigint): bigint {
-        return MathSol.divDownFixed(amount, this.amount + WAD);
+    /**
+     * Applies slippage to an amount in a given direction
+     *
+     * @param amount amout to apply slippage to
+     * @param direction +1 adds the slippage to the amount, and -1 will remove the slippage from the amount
+     * @returns
+     */
+    public applyTo(amount: bigint, direction = 1): bigint {
```
brunoguerios, line 50: We're striving to make the b-sdk interface as explicit as possible without the need to read through docs/comments to know how to interact with any function.
I think this `direction` parameter goes against this idea as the user would have to infer/read that `>=0` is to add slippage and `< 0` is to remove.
I believe the best way to handle this is by making the type more strict, such as: `1 | -1`
Since it's a bit of a cosmetic thing, I'll apply changes on my end, so you don't have to spend more time on this.
Hope it's ok 😉

PR #233, opened 2024-02-02
Basic Remove Liquidity for V3

### LINKED ISSUE #218: V3: Basic Remove Liquidity

brunoguerios:
closes #218 

Main changes:
- Add remove liquidity for v3

Other changes:
- Rename RemoveLiquidityKind.SingleToken to RemoveLiquidityKind.SingleTokenExactIn
- Add RemoveLiquidityKind.SingleTokenExactOut for v2 as well using Unbalanced under the hood


src/entities/removeLiquidity/removeLiquidityV3/index.ts
```diff
@@ -0,0 +1,228 @@
...
+                    tokenOutIndex = undefined;
+                }
+                break;
+        }
+
+        const output: RemoveLiquidityBaseQueryOutput = {
+            poolType: poolState.type,
+            removeLiquidityKind: input.kind,
+            poolId: poolState.id,
+            bptIn,
```
johngrantuk, line 139: This is showing some intellisense errors for me here and line below.

src/entities/removeLiquidity/removeLiquidityV3/index.ts
```diff
@@ -0,0 +1,228 @@
...
+
+        let bptIn: TokenAmount;
+        let amountsOut: TokenAmount[];
+        let tokenOutIndex: number | undefined;
+
+        switch (input.kind) {
+            case RemoveLiquidityKind.Unbalanced:
+                throw removeLiquidityUnbalancedNotSupportedOnV3;
+            case RemoveLiquidityKind.SingleTokenExactOut:
+                {
```
johngrantuk, line 48: Is it possible to make use of `getAmountsQuery` helper?

src/entities/removeLiquidity/removeLiquidityV3/index.ts
brunoguerios: Thanks for pointing that out. It's missing the RemoveLiquidityRecovery implementation. I'll add it over here.

src/entities/removeLiquidity/removeLiquidityV3/index.ts
brunoguerios: I believe so 👍  I'll refactor over here.

src/entities/removeLiquidity/removeLiquidityV3/index.ts
brunoguerios: Double checked and it's pending SC implementation and Router ABI updates. I'll throw for now and create a task to keep track of it.

src/entities/removeLiquidity/removeLiquidityV3/index.ts
brunoguerios: done ✅

brunoguerios:
I added an extra refactor to improve readability - let me know what you think. Thanks!

src/entities/removeLiquidity/helper.ts
```diff
@@ -0,0 +1,83 @@
...
+            return {
+                minAmountsOut: getAmounts(tokens, [input.amountOut]),
+                tokenOutIndex: tokens.findIndex((t) =>
+                    t.isSameAddress(input.amountOut.address),
+                ),
+                maxBptAmountIn: MAX_UINT256, // maxAmountIn set to max when querying
+            };
+        case RemoveLiquidityKind.SingleTokenExactIn:
+            return {
+                minAmountsOut: Array(tokens.length).fill(1n), // minAmountsOut set to 1 wei when querying
```
johngrantuk, line 35: Is this right?

johngrantuk (approved):
Apart from that last comment in `RemoveLiquidityKind.SingleTokenExactIn` it LGTM,

src/entities/removeLiquidity/helper.ts
brunoguerios: Yeah, that was intentional. But we don't need to worry too much about it.
Giving a bit more context, on v3, SC expects minAmountsOut to be non-zero. 
Specially on the SingleTokenExactOut case, where they expect minAmountsOut to have a single non-zero value, at the index of the tokenOut.
That's how they identify which token to remove liquidity to.
But, having said that, the plan is to remove minAmountsOut from query inputs as a whole, as they don't see value on having it there.
So we'll have to update this when the new Router is released.
Note: on the actual transaction call, they will enforce minAmountsOut to be non-zero

PR #234, opened 2024-02-05
Adding bptOut to the calculation of the proportional amounts;

lgahdl:
Changing example pool to Gyroscope ECLP wstETH/wETH;

### DIFF
--- a/examples/utils/calculateProportionalAmounts.ts
+++ b/examples/utils/calculateProportionalAmounts.ts
@@ -11,18 +11,23 @@ export default async function calculateProportionalAmountsExample() {
         chainId,
     );
     const poolId =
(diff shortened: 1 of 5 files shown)


brunoguerios (approved):
That looks nice 👏 
Did you double check with other pool types?
Do you think it's worth an integration test comparing the calculated results with the output of a Remove Liquidity Proportional?

lgahdl:
> That looks nice 👏 Did you double check with other pool types? Do you think it's worth an integration test comparing the calculated results with the output of a Remove Liquidity Proportional?

No, good point, I need to check for GyroE V1, Gyro2 and Gyro3.
About the tests, I don't think it's worth to have integration tests for this utility function, since it's just basic math, and not a function with direct integration with the contracts, but I do think unit tests would be good to have though

brunoguerios:
Awesome that it also works to calculate bptOut 🙌

PR #237, opened 2024-02-06
Swap refactor for use with API SOR

johngrantuk:
Refactor `Swap` to interface with response from API.

Assume its still important that `Swap` returns TokenAmounts to consumer rather than bigint to improve dev ex. API can't return a TokenAmount so should return a raw amount instead (e.g. 1USDC=1000000). If it also returns decimals information the b-sdk/Swap object can create a Token so the user doesn't have to.

Suggested API response:
```
type GqlSorPath {
   pools: [String]! --- note this would be address?
   tokens: [Token]!
   outputAmountRaw: String!
   inputAmountRaw: String!
}

type Token {
    address: String!
    decimals: Int!
}
```

Pseudo code for user flow:
```
const sorPaths = api.getSwapPaths(swapKindd, tokenIn, tokenOut, ...); // Returns array of GqlSorPaths

const swap = new Swap(chainId, sorPaths, swapKind);

const apiQuote = swap.quote();

const updated = await swap.query();
```

Proposed API response returns an array of `SorPath` objects which consumer can use as inputs to the constructor of `Swap`.
Swap uses `Path[]` to create `PathWithAmount[]` and `PathWithAmount` creates Token/TokenAmount.  

Note - Also need to add Limits, Call construction etc to Swap as in this PR: https://github.com/balancer/b-sdk/pull/220

### DIFF
--- a/src/entities/swapRefactorDraft.ts
+++ b/src/entities/swapRefactorDraft.ts
@@ -1,0 +1,376 @@
+import { TokenAmount } from './tokenAmount';
+import { SingleSwap, SwapKind, BatchSwapStep } from '../types';
+import {
+    abs,
+    BALANCER_QUERIES,
+    DEFAULT_USERDATA,
+    DEFAULT_FUND_MANAGMENT,
+    ZERO_ADDRESS,
+    NATIVE_ADDRESS,
+    MathSol,
+} from '../utils';
+import {
+    Address,
+    createPublicClient,
+    encodeFunctionData,
+    getContract,
+    http,
+} from 'viem';
+import { balancerQueriesAbi } from '../abi';
+import { PriceImpactAmount } from './priceImpactAmount';
+import cloneDeep from 'lodash.clonedeep';
+import { Token } from './token';
+import { MinimalToken } from '..';
+
+/**
+API Returns list of GqlSorPath:
+
+type GqlSorPath {
+   pools: [String]! --- note can this be address?
+   tokens: [Token]!
+   outputAmountRaw: String!
+   inputAmountRaw: String!
+}
+
+type Token {
+    address: String!
+    decimals: Int!
+}
+*/
+
+type TokenApi = Omit<MinimalToken, 'index'>;
+
+type Path = {
+    pools: Address[];
+    tokens: TokenApi[];
+    outputAmountRaw: bigint;
+    inputAmountRaw: bigint;
+};
+
+class PathWithAmount {
+    public readonly pools: Address[];
+    public readonly tokens: TokenApi[];
+    public readonly outputAmount: TokenAmount;
+    public readonly inputAmount: TokenAmount;
+
+    public constructor(
+        chainId: number,
+        tokens: TokenApi[],
+        pools: Address[],
+        inputAmountRaw: bigint,
+        outputAmountRaw: bigint,
+    ) {
+        if (pools.length === 0 || tokens.length < 2) {
+            throw new Error(
+                'Invalid path: must contain at least 1 pool and 2 tokens.',
+            );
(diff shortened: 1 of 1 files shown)


src/entities/swapRefactorDraft.ts
```diff
@@ -0,0 +1,376 @@
...
+import { PriceImpactAmount } from './priceImpactAmount';
+import cloneDeep from 'lodash.clonedeep';
+import { Token } from './token';
+import { MinimalToken } from '..';
+
+/**
+API Returns list of GqlSorPath:
+
+type GqlSorPath {
+   pools: [String]! --- note can this be address?
```
brunoguerios, line 29: Considering this needs to support both v2 and v3, I believe we'll have to stick with id instead of address. For v3 api will copy pool address into pool id, so we should be covered.

src/entities/swapRefactorDraft.ts
```diff
@@ -0,0 +1,376 @@
...
+                    chainId,
+                    p.tokens,
+                    p.pools,
+                    p.inputAmountRaw,
+                    p.outputAmountRaw,
+                ),
+        );
+
+        // paths with immutable pool balances
+        this.pathsImmutable = cloneDeep(this.paths);
```
brunoguerios, line 116: This was needed in the context of the previous PathWithAmount that had mutating balances. Since this is no longer the case, you should be able to remove it.

src/entities/swapRefactorDraft.ts
```diff
@@ -0,0 +1,376 @@
...
+    public get priceImpact(): PriceImpactAmount {
+        const paths = this.pathsImmutable;
+
+        const pathsReverse = paths.map(
+            (path) =>
+                new PathWithAmount(
+                    this.chainId,
+                    [...path.tokens].reverse(),
+                    [...path.pools].reverse(),
+                    path.outputAmount.amount, // TODO - Is this correct scale?
```
brunoguerios, line 257: Yes, since you're using `TokenAmount.fromRawAmount()`, this seems to be the correct scaling 👍

src/entities/swapRefactorDraft.ts
```diff
@@ -0,0 +1,376 @@
...
+
+    private getSwaps(paths: PathWithAmount[]) {
+        let swaps: BatchSwapStep[] | SingleSwap;
+        if (this.isBatchSwap) {
+            swaps = [] as BatchSwapStep[];
+            if (this.swapKind === SwapKind.GivenIn) {
+                paths.map((p) => {
+                    p.pools.map((pool, i) => {
+                        (swaps as BatchSwapStep[]).push({
+                            poolId: pool, // TODO - Expect ID but currently using Address
```
brunoguerios, line 291: As I said above, I believe we should expect ids from the api and use it like that.

src/entities/swapRefactorDraft.ts
```diff
@@ -0,0 +1,376 @@
...
+            );
+            swaps = {
+                poolId: pool, // TODO - Expect ID but currently using Address
+                kind: this.swapKind,
+                assetIn,
+                assetOut,
+                amount:
+                    this.swapKind === SwapKind.GivenIn // TODO - Double check
+                        ? path.inputAmount.amount
+                        : path.outputAmount.amount,
```
brunoguerios, line 342: This seems correct to me. What do you need to double check?

brunoguerios (commented):
I believe this is totally going in the right direction 👏 
Awesome that we're able to simplify the interface and have it in a way that the FE can easily consume it 🚀

johngrantuk:
Closing in favour of an alternative branch.

PR #238, opened 2024-02-07
Expose JoinPoolRequest type

brunoguerios:


brunoguerios:
Since this is quite a small change I'll go ahead and merge it so it can be shipped with today's release.

PR #239, opened 2024-02-07
Swap api refactor

johngrantuk:
Refactor `Swap` to interface with response from API.

- Added `fetchSorSwapPaths` to `BalancerApi` as helper to fetch paths from API easily.
- Swap class has V2/V3 handler  (V3 WIP in follow up PR)
- SwapV2 refactored to take API response (see below for details)
- [Swap example](https://github.com/balancer/b-sdk/blob/swap-api-refactor/examples/swap.ts) added
- Limits refactored to be internal to V2 implementation. `buildCall` used to create based on input Slippage. Returns `minAmountOut`/`maxAmountIn` for users.
- PriceImpact can no longer be calculated in `Swap` as missing pool state/maths. This has been removed for now and will be added as onchain helper in future.

### Implementation Notes

Suggested API response:
```
API Returns list of GqlSorPath:

type GqlSorPath {
  balancerVersion: Int!
  pools: [String]!
  tokens: [Token]!
  outputAmountRaw: String!
  inputAmountRaw: String!
}

type Token {
    address: String!
    decimals: Int!
}
```

See [swap example](https://github.com/balancer/b-sdk/blob/swap-api-refactor/examples/swap.ts) for user flow:
```
const chainId = ChainId.MAINNET;

// API is used to fetch best path from available liquidity
const balancerApi = new BalancerApi(
    'https://backend-v3-canary.beets-ftm-node.com/graphql',
    chainId,
);

const sorPaths = await balancerApi.sorGetQuote.fetchSorGetQuote(....inputParams);

// Swap object provides useful helpers for re-querying, building call, etc
const swap = new Swap({
    chainId,
    paths: sorPaths,
    swapKind: SwapKind.GivenIn,
});

console.log(
    `Input token: ${swap.inputAmount.token.address}, Amount: ${swap.inputAmount.amount}`,
);
console.log(
    `Output token: ${swap.outputAmount.token.address}, Amount: ${swap.outputAmount.amount}`,
);
console.log(`Price Impact: ${swap.priceImpact.percentage}%`);

// Get up to date swap result by querying onchain
const updated = await swap.query(process.env.ETHEREUM_RPC_URL);
console.log(updated.amount);
```

Assume its still important that `Swap` returns TokenAmounts to consumer rather than bigint to improve dev ex. API can't return a TokenAmount so should return a raw amount instead (e.g. 1USDC=1000000). If it also returns decimals information the b-sdk/Swap object can create a Token so the user doesn't have to.

Proposed API response returns an array of `SorPath` objects which consumer can use as inputs to the constructor of `Swap`.
Swap uses `Path[]` to create `PathWithAmount[]` and `PathWithAmount` creates Token/TokenAmount.


src/entities/swap/swapV2/index.ts
```diff
@@ -0,0 +1,339 @@
...
+                    chainId,
+                    p.tokens,
+                    p.pools,
+                    p.inputAmountRaw,
+                    p.outputAmountRaw,
+                ),
+        );
+
+        // paths with immutable pool balances
+        this.pathsImmutable = cloneDeep(this.paths);
```
brunoguerios, line 47: I feel like we won't need this pathsImmutable. You could also remove cloneDeep dependency as I believe this is the only place we use it.

brunoguerios (commented):
Nice V3 split and refactor 👍 
Do you think it's worth to have a Swap interface that SwapV2 and SwapV3 implement?

johngrantuk:
> Nice V3 split and refactor 👍 Do you think it's worth to have a Swap interface that SwapV2 and SwapV3 implement?

Yeah, I was thinking the same, I'll add the interface. Also plan on adding some specific `Swap` tests too.

src/entities/swap/types.ts
```diff
@@ -0,0 +1,64 @@
...
+    swaps: BatchSwapStep[] | SingleSwap;
+    quote: TokenAmount;
+    inputAmount: TokenAmount;
+    outputAmount: TokenAmount;
+    query(rpcUrl?: string, block?: bigint): Promise<TokenAmount>;
+    queryCallData(): string;
+    buildCall(swapCall: SwapCallBuild): SwapBuildOutputBase;
+}
+
+type BaseSwapCall = {
```
brunoguerios, line 46: A small suggestion for consistency:
```suggestion
type SwapCallBase = {
```

test/entities/swap.test.ts
```diff
@@ -1,126 +1,229 @@
...
+        describe('GivenIn', () => {
+            test('18decimals>6decimals: minAmountOut to be 0.1% less then expected', () => {
+                const swap = new Swap({
+                    chainId: ChainId.MAINNET,
+                    paths: [pathTo6Decimals],
+                    swapKind: SwapKind.GivenIn,
+                });
+                const slippage = Slippage.fromPercentage('0.1');
+                const tokenOut = new Token(
+                    1,
```
brunoguerios, line 111: To avoid using incorrect ids, we we could create a `chainId` variable with a broader scope and reuse. What do you think?

brunoguerios (approved):
I added some small suggestions, but otherwise it looks great 👏

PR #240, opened 2024-02-08
Remove SOR related code & other stale code/exports

johngrantuk:
- Tests to sanity check:
  - test/entities/swap.test.ts
  - test/v2/priceImpact/priceImpact.sor.integration.test.ts (WIP - needs updated once confirmed aim of tests here)
- Used `npx ts-unused-exports tsconfig.json` to find other stale code/exports and removed


brunoguerios (approved):
That's indeed a lot of code removal 😅

PR #241, opened 2024-02-09
Use macos14 (M1) in GitHub ations

agualis:


agualis:
It was not working our of the box so I will investigate when I have more time.

PR #244, opened 2024-02-09
Add price impact for add/remove nested liquidity

### LINKED ISSUE #134: Price Impact - ABA method - Nested Liquidity
Implement price impact for nested liquidity within b-sdk.
Note: releasing this depends on relayerV6 being audited and approved by governance. Development can still be done with a locally deployed relayer. It should be ok to build on top of the existing nested liquidity working branch.

- [ ] add liquidity nested unbalanced
- [ ] remove liquidity nested single token

brunoguerios:
Closes #134 and #177 

Main:
- Add price impact for Remove Liquidity Nested (single token)
- Add price impact for Add Liquidity Nested (unbalanced)

Extra:
- Add alternative method for Add Liquidity Unbalanced. Since it relies on Remove Liquidity Unbalanced, it works on v2 only. I thought it would be interesting to have it in the repo because it serves the purpose of validating the more generic price impact calculation for Add Liquidity Unbalanced, which is the only one that uses a modified ABA approach.
- Add comments describing each price impact method
- Refactor some constants on addLiquidityNested tests

I added a lot of info to the comments. Maybe more than necessary making it too verbose?
I'm open to feedback/suggestions on those as well 😉 

Cheers!


test/v2/priceImpact/priceImpact.integration.test.ts
```diff
@@ -18,34 +18,50 @@ import {
     RemoveLiquidityKind,
     RemoveLiquidityUnbalancedInput,
     PoolType,
+    RemoveLiquidityNestedSingleTokenInput,
+    NestedPoolState,
+    ZERO_ADDRESS,
+    AddLiquidityNestedInput,
 } from '../../../src';
 import { ANVIL_NETWORKS, startFork } from '../../anvil/anvil-global-setup';
 import { PriceImpact } from '../../../src/entities/priceImpact';
```
johngrantuk, line 27: Could import from `src` to keep consistent and to make sure they are exported correctly.

test/v2/priceImpact/priceImpact.integration.test.ts
```diff
@@ -117,16 +136,62 @@ describe('price impact', () => {
...
+                ],
+            };
+        });
+        test('ABA close to Spot Price', async () => {
+            const priceImpactABA = await PriceImpact.addLiquidityNested(
+                input,
+                nestedPoolState,
+            );
+            const priceImpactSpot = PriceImpactAmount.fromDecimal(
+                '0.0472', // TODO: find a way to validate this result
```
johngrantuk, line 176: Something you've checked with Zen?

src/entities/priceImpact/index.ts
```diff
@@ -200,6 +218,142 @@ export class PriceImpact {
...
+            amountsOut: amountsIn.map((a) => a.toInputAmount()),
+            kind: RemoveLiquidityKind.Unbalanced,
+        };
+        const { bptIn } = await removeLiquidity.query(
+            removeLiquidityInput,
+            poolState,
+        );
+
+        // get relevant amounts for price impact calculation
+        const amountInitial = parseFloat(formatUnits(bptOut.amount, 18));
```
johngrantuk, line 261: Is there a reason to use numbers here but bigint in `singleSwap`? Could maybe make a single helper ABA method on `PriceImpact` class that all could call and use consistent type?

src/entities/priceImpact/index.ts
```diff
@@ -200,6 +218,142 @@ export class PriceImpact {
...
 
+    /**
+     * Alternative for calculating price impact on add liquidity unbalanced operations
+     *
+     * Note: works on balancer v2 only because it relies on Remove Liquidity Unbalanced
+     * @param input same input used in the corresponding add liquidity operation
+     * @param poolState same pool state used in the corresponding add liquidity operation
+     * @returns price impact amount
+     */
+    static addLiquidityUnbalancedAlternative = async (
```
johngrantuk, line 229: When would someone choose addLiquidityUnbalancedAlternative over addLiquidityUnbalanced? Do we need to keep both?

src/entities/priceImpact/index.ts
johngrantuk, line 1: Is there a TLDR description of what the addLiquidityUnbalanced unbalanced method is? e.g. explaing deltas, etc

src/entities/priceImpact/index.ts
```diff
@@ -237,6 +391,59 @@ export class PriceImpact {
...
+        // calculate price impact using ABA method
+        const priceImpact = (amountInitial - amountFinal) / amountInitial / 2;
+        return PriceImpactAmount.fromDecimal(`${priceImpact}`);
+    };
+
+    /**
+     * Calculate price impact on single swap operations
+     * @param input same input used in the corresponding single swap operation
+     * @returns price impact amount
+     */
```
johngrantuk, line 446: `singleSwap` won't work with V3. Could construct Path and use `Swap` class which should expose the query for both V2 and V3.

johngrantuk (commented):
Some of the functions end up doing a lot of onchain/rpc calls. Any idea how it would affect performance/UX on FE? Could add the PI to the examples which would give us an idea when we run locally and might highlight any issues with the flow.

test/v2/priceImpact/priceImpact.integration.test.ts
brunoguerios: Yeah, I'm actually using the old SDK results for comparison. I left this comment so I wouldn't forget to do it. Thanks for bringing my attention to it.
Ps: Zen was helping on validating the approach in a more generalized way, coming up with relevant scenarios for comparison and analyzing the results. Now that it's validated, it's easier to compare this scenario directly with results from the old SDK.

src/entities/priceImpact/index.ts
brunoguerios: Yeah, you're right. A helper method does make sense for this. I'll do it over here.
For context, I wasn't worrying too much about numbers vs bigint precision because price impact is meant to be an approximation. But I agree that we should be consistent anyway. 👍

src/entities/priceImpact/index.ts
brunoguerios: I added the reason to the PR description:
> I thought it would be interesting to have it in the repo because it serves the purpose of validating the more generic price impact calculation for Add Liquidity Unbalanced, which is the only one that uses a modified ABA approach.

This alternative method actually follows more precisely the principle of the ABA method, but is only available for v2.
The main one is an adaptation that works with both v2 and v3 because it doesn't rely on Remove Liquidity Unbalanced.

I can see that it may cause confusion, and that's why I added comments explaining where they can be applied.
Let me know if you still think we're better off without it and I'll remove it over here.

johngrantuk: I wasn't really sure who would find it useful for validation? - If its just us internally then I think we should move it out to a test helper (assuming validation is during testing?). If we're not using it for testing then I'd remove it completely, could always copy the snippet to Notion if you think it'll be useful in future.

For the consumer I definitely think it'll be confusing which to use - if we have something that can cover V2 & V3 we should just use that by default and take the decision away 👍.

src/entities/priceImpact/index.ts
brunoguerios: Yes, I documented the steps on [this Notion doc](https://www.notion.so/Price-Impact-49545b95b98243e3ae624be6bcea2700?pvs=4#419a8635787a4c7db14528198495d46a).
The first block has a more brief overview of the approach and below you'll find a more detailed step-by-step description.
I agree it makes sense to have that in the code as well, so it's easier to follow what's going on.
I'll add it there so you can provide feedback.

src/entities/priceImpact/index.ts
brunoguerios: I'm wondering if I should remove method as a whole.
The initial goal was to have the swap class using this helper internally, but that doesn't really make much sense.
If we already have the swap class providing price impact for a swap, this feels a bit redundant.
What do you think?

johngrantuk: The Swap class no longer has a PI implementation because we removed the maths so this was quite nice. I did a follow up PR [here](https://github.com/balancer/b-sdk/pull/252) to show how it could be used.

brunoguerios:
I agree it's really nice that it doesn't rely on pool math 😃 
About the performance concern, you're right. Most methods rely on only 2 queries for calculating price impact, which should be ok. The only one that could be a problem is the add liquidity unbalanced and we already have a task to work on a helper contract. That will be my main focus next week (and likely the next one).
I'll add examples to make sure we have everything working as expected. 👍 
And nice that you like the comments/docs - I'll add more for add liquidity unbalanced ✅

src/entities/priceImpact/index.ts
johngrantuk: That Notion doc is super helpful/well-written 🙏.

src/entities/priceImpact/index.ts
brunoguerios: I agree that it's confusing for the consumer and that it's beneficial to take the decision away 👍 
I like the idea of having within a test. I'll do it over here.

src/entities/priceImpact/index.ts
brunoguerios: I'm actually thinking of a way to have Price Impact for swaps being handled externally (the same way as add/remove liquidity) so we don't need to worry about v2/v3 or single/batch swaps.
Ideally we'd provide it with a `SwapInput` and it would be able to build both path and reverse path, query amounts and return price impact.
I'll create a separate tasks so we don't forget to take a look at this in the future.
Let me know what you think of the approach.

brunoguerios:
I handled most of the comments so far, except for examples, which should be done hopefully by tomorrow. Feel free to review the progress so far, as it's mostly independent from what's pending.

src/entities/priceImpact/index.ts
johngrantuk: You mean an onchain helper type solution that we can make with a single call? I think that would be the ideal solution.

brunoguerios: I agree that this would be the ideal end goal, but I'd rather start with an offchain solution first (that we can improve/fix if necessary) and move to a helper contract only after it's battle tested.
Since the solution itself would require only 2 onchain calls (like price impact for add/remove liquidity) it should be ok.

johngrantuk: Yeah, sounds good. So just to be clear - for now, we're going to keep an offchain method on PriceImpact which will support v2 and v3?

lgahdl:
Except for the conflicts, LGTM! 👍🏻

src/entities/priceImpact/index.ts
brunoguerios: Yes 👍

examples/priceImpact/addLiquidity.ts
johngrantuk, line 1: Feels like it would make more sense to add the price impact to the existing add/remove liquidity examples rather than having separate ones? Thinking that its usually going to be part of that flow and makes it clearer/easier for users to see.

brunoguerios: I thought of that possibility. 
I agree that it feels more natural to look at price impact within the user flow of add/remove liquidity.
But on the other hand, as a dev, I'd look for priceImpact examples directly to see how it's used.
Maybe we add price impact to each add/remove liquidity example and also keep one separated to be used as reference?

johngrantuk: Yeah that sounds nice to me. Could add some comments on the separate one highlighting there are different options.

brunoguerios: done ✅

PR #245, opened 2024-02-09
V3 - Create Weighted Pool

lgahdl:
Adds Balancer's V3 CreatePool functionality for Weighted Pools;
Adds tests for the V3's Create Weighted Pool;
Adds example for V3's Create Weighted Pool;
Renames factory references to differ by balancer version instead of pool version, or no differentiation;
Adapts createPoolHelper to work with V3;
Adds TokenType enum;
Add another input validation step for V3's Create Weighted Pool input("rateProvider" cannot be zero if "tokenType" is not STANDARD);


examples/v3/createPoolWeighted.ts
```diff
@@ -0,0 +1,78 @@
+import dotenv from 'dotenv';
+dotenv.config();
+
+import { CreatePool, CreatePoolV3WeightedInput } from '@/entities';
```
johngrantuk, line 4: We want to have all the non-test related code imported from the main src so the user can basically copy and paste.

examples/v3/createPoolWeighted.ts
```diff
@@ -0,0 +1,78 @@
...
+                weight: parseEther(`${1 / 2}`),
+                rateProvider: zeroAddress,
+                tokenType: TokenType.STANDARD,
+            },
+        ],
+        balancerVersion: 3,
+    };
+    const { call } = createPool.buildCall(createWeightedPoolInput);
+    const hash = await client.sendTransaction({
+        to: WEIGHTED_POOL_FACTORY_BALANCER_V3[chainId],
```
johngrantuk, line 55: Would it be better to return this as part of the buildCall response?

test/v3/createPool/weighted.integration.test.ts
```diff
@@ -0,0 +1,172 @@
...
+    });
+    test('Create Weighted Pool', async () => {
+        poolAddress = await doCreatePool({
+            ...txInput,
+            createPoolInput: createWeightedPoolInput,
+        });
+        expect(poolAddress).to.not.be.undefined;
+    });
+
+    test('Wrong weights, expects error', async () => {
```
johngrantuk, line 75: Should the following tests be moved to non-integration?

src/entities/createPool/createPoolV3/index.ts
```diff
@@ -0,0 +1,29 @@
+import { PoolType } from '@/types';
+import {
+    CreatePoolBase,
+    CreatePoolBuildCallOutput,
+    CreatePoolInput,
+} from '../types';
+import { CreatePoolWeightedV3 } from './weighted/createPoolWeighted';
+
+export class CreatePoolV3 implements CreatePoolBase {
+    // TODO: should we allow for custom create pool types?
```
johngrantuk, line 10: I'm not sure if this is important but might be worth bringing this up with Daniel?

src/entities/createPool/createPoolV3/weighted/createPoolWeighted.ts
```diff
@@ -0,0 +1,64 @@
...
+            functionName: 'create',
+            args,
+        });
+        return { call: encodedCall };
+    }
+
+    private parseCreateFunctionArgs(
+        input: CreatePoolV3WeightedInput,
+    ): CreatePoolV3WeightedArgs {
+        const sortedTokenParams = input.tokens.sort(
```
johngrantuk, line 26: I wonder if something along the lines of [this](https://github.com/balancer/balancer-v3-monorepo/blob/9d1dd909555309f5c59a504c2dca7fc6018fea8a/pvt/helpers/src/models/tokens/ERC20TokenList.ts#L168) is easier to follow? 
We can probably break the sorting to a separate helper function too and re-use as the same is being done in each pool type creation.

johngrantuk (changes requested):
Added a few comments but otherwise looks pretty good to me.

test/v3/createPool/weighted.integration.test.ts
lgahdl: Absolutely, thanks! I created a task on the board to also verify and fix this on the V2 Create and Init Pool test files

src/entities/createPool/createPoolV3/weighted/createPoolWeighted.ts
lgahdl: Actually I already did a function to sort tokens by address, I forgot to reuse 😅

src/entities/createPool/createPoolV3/index.ts
lgahdl: I think it's impossible to do so 😕 Because the CreatePoolInput would also be custom depending on the pool type, we cannot typify the buildCall input for custom Create Pool Types, the AddLiquidity and RemoveLiquidity can have custom types because the input type of the function doesn't change.

brunoguerios: Yeap. I added this comment a while ago to investigate, but we already discussed and reached a conclusion that this is not possible. You can remove the comment now.

lgahdl: I was thinking about it today, and maybe we can handle this by using generic types if we need.
I'll ask Daniel if this is an interesting feature anyway.

.changeset/unlucky-ads-film.md
```diff
@@ -0,0 +1,11 @@
...
+"@balancer/sdk": minor
+---
+nt
+Adds Balancer's V3 CreatePool functionality for Weighted Pools;
+Adds tests for the V3's Create Weighted Pool;
+Adds example for V3's Create Weighted Pool;
+Renames factories references to differ by balancer version instead of pool version, or no differentiation;
+Adapts createPoolHelper to work with V3;
+Adds TokenType enum;
+Add another validation step for V3's Create Weighted Pool input(rateProvider cannot be zero if tokenType is not STANDARD);
```
brunoguerios, line 11: These are actually release notes, so you could/should be more concise.
> Add V3 CreatePool functionality for Weighted Pools

This one likely covers all items on this list.
If you really want to highlight something else, feel free to add another line or two, otherwise you can remove most of them.

examples/v3/createPoolWeighted.ts
```diff
@@ -0,0 +1,78 @@
...
+    });
+    const transactionReceipt = await client.waitForTransactionReceipt({
+        hash,
+    });
+
+    const poolCreatedEvent = findEventInReceiptLogs({
+        receipt: transactionReceipt,
+        eventName: 'PoolCreated',
+        abi: weightedPoolFactoryV3Abi,
+        to: WEIGHTED_POOL_FACTORY_BALANCER_V3[chainId],
```
brunoguerios, line 68: And reuse it here.

test/v3/createPool/weighted.integration.test.ts
```diff
@@ -0,0 +1,75 @@
...
+import {
+    Address,
+    createTestClient,
+    http,
+    parseEther,
+    publicActions,
+    walletActions,
+    zeroAddress,
+} from 'viem';
+import { CHAINS, ChainId, PoolType, TokenType } from '../../../src';
```
brunoguerios, line 12: You should import directly from `src` for consistency and to make sure everything is properly exported.

src/abi/composableStableFactory.V2.ts
```diff
@@ -1,4 +1,4 @@
-export const composableStableFactoryV5Abi = [
+export const composableStableFactoryV2Abi = [
```
brunoguerios, line 1: I think it's still nice to mention the factory version as well, since this might change in the future.
Here are a few options for us to evaluate:
- `composableStableFactoryV5Abi_V2`
- `composableStableFactoryV5Abi_balancerV2`
- `composableStableFactoryV5Abi_vaultV2`

brunoguerios (commented):
Here are a few extra comments/suggestions

examples/v3/createPoolWeighted.ts
brunoguerios: I see now that you're returning it on the function, so all you need is to use it on the example 😉

src/abi/index.ts
```diff
@@ -3,7 +3,13 @@ export * from './balancerQueries';
 export * from './balancerRelayer';
 export * from './balancerRouter';
 export * from './batchRelayerLibrary';
+export * from './composableStableFactoryV5.V2';
+export * from './composableStablePoolV5.V2';
 export * from './erc20';
 export * from './vaultV2';
 export * from './vaultV3';
-export * from './weightedFactoryV4';
+export * from './vaultV3Extension';
```
brunoguerios, line 11: If you look at vaultV3 abi, you'll see that it's already taking into account the extension. We should always access the extension through the main vault, so exposing this abi separately is not useful and my generate confusion.
```suggestion
```

brunoguerios (commented):
I left one final suggestion, otherwise LGTM 👍 
I'll leave the approval for @johngrantuk in case he wants to take a final look.

examples/v2/createPoolWeighted.ts
```diff
@@ -34,23 +35,23 @@ const createPool = async (stopForkAfterExecution = true) => {
     const signerAddress = (await client.getAddresses())[0];
     const createPool = new CreatePool();
     const poolType = PoolType.Weighted;
-    const createWeightedPoolInput: CreatePoolWeightedInput = {
+    const createWeightedPoolInput: CreatePoolV2WeightedInput = {
```
johngrantuk, line 38: Shows a warning for me: `'chainId' is declared here`

examples/v2/createPoolComposableStable.ts
```diff
@@ -1,29 +1,27 @@
 import dotenv from 'dotenv';
 dotenv.config();
 
+import {
+    composableStableFactoryV5Abi_V2,
+    CreatePool,
+    CreatePoolV2ComposableStableInput,
+} from 'src';
+import { PoolType } from '@/types';
+import { ChainId, CHAINS } from '@/utils';
```
johngrantuk, line 10: Some of these can still be moved to be imported from src

src/utils/sortByAddress.ts
```diff
@@ -0,0 +1,13 @@
...
+
+interface ObjectWithAddress {
+    address: Address;
+}
+// Sorts any array of objects by an 'address' param
+export function sortByAddress<T extends ObjectWithAddress>(
+    objectWithAddressArray: T[],
+): T[] {
+    return objectWithAddressArray.sort((a, b) =>
+        a.address.localeCompare(b.address),
```
johngrantuk, line 11: I'm not familiar with localCompare - are we sure it sorts as we expect?

johngrantuk (changes requested):
Looks good. Just a few last comments/suggestions.

src/utils/sortByAddress.ts
lgahdl: This function compares strings using the Unicode charset:

`0, 1, 2, ... 9, A, B, C, D, ... Z, a, b, c, ..., z`

 It works fine if both strings are lowercase, but we would not have the correct result comparing a "0xE5..." with a "0xe4..." address, thanks for noticing, I'll add an extra step to convert both strings to lowercase

examples/v2/createPoolWeighted.ts
lgahdl: Done

johngrantuk (approved):

PR #246, opened 2024-02-12
Add integration test for Nested Add Liquidity in Polygon

agualis:
This PR adds an integration test that throws an error when trying to add liquidity in [this nested pool in polygon](https://v3.balancer.fi/pools/polygon/v2/0x60f46b189736c0d2ae52a79382b64c1e2a86b0d9000200000000000000000cc4).


brunoguerios:
Thanks for reporting this @agualis !
b-sdk was indeed not supporting any relayer operations on chains other than mainnet.
I fixed this and some other minor things that will help us spot some missing test configs.

johngrantuk (approved):

PR #250, opened 2024-02-15
V3 Single Swap support

johngrantuk:
* batchSwap to be done in future PR
* Note change to `Swap` constructor params to handle different V2/V3 options, sender/recipient, wethIsEth. Not sure if this is best approach and looking for feedback.


test/entities/swapV3.integration.test.ts
```diff
@@ -0,0 +1,335 @@
...
+                        wethIsEth: true,
+                    });
+                    await assertSwapExactOut(client, chainId, swap, true);
+                });
+            });
+        });
+    });
+});
+
+async function assertSwapExactIn(
```
lgahdl, line 202: Would be good to move this function to a different file on the folder `./test/lib/utils`, just as another test helper functions.

test/entities/swapV3.integration.test.ts
```diff
@@ -0,0 +1,335 @@
...
+    }
+
+    expect(balanceDeltas).to.deep.eq([
+        expectedEthDelta,
+        expectedTokenInDelta,
+        expectedTokenOutDelta,
+    ]);
+}
+
+async function assertSwapExactOut(
```
lgahdl, line 269: Same with this one

test/entities/swapV3.integration.test.ts
johngrantuk: I think its better to keep it here for now as its not being used anywhere else. Might change after batchSwaps are added though.

src/entities/swap/index.ts
```diff
@@ -2,36 +2,31 @@ import { TokenAmount } from '../tokenAmount';
...
+    public constructor(swapInput: SwapInput) {
+        validatePaths(swapInput.paths);
 
-        switch (paths[0].balancerVersion) {
-            case 2:
-                this.swap = new SwapV2({ chainId, paths, swapKind });
-                return;
-            case 3:
-                throw new Error('Unsupported Balancer Protocol Version');
+        if ('wethIsEth' in swapInput) {
```
brunoguerios, line 26: On Add/Remove Liquidity implementation we have the same external interface for both v2 and v3, where we expose a flag that the user specifies if the input is ETH or not.
Currently `useNativeAssetAsWrappedAmountIn`, but soon to be refactored to be consistent with `wethIsEth` from SC.
Internally we check that flag and change addresses to ADDRESS_ZERO accordingly.
If we follow the same pattern, we should ideally expose the same flag for swapV2 and update addresses internally.
Checking for `wethIsEth` flag wouldn't work on this scenario.
A suggestion would be to use `balancerVersion` available within `swapInput.paths[0]` instead.

examples/swap.ts
```diff
@@ -64,6 +63,8 @@ const swap = async () => {
         chainId,
         paths: sorPaths,
         swapKind,
+        sender,
+        recipient,
```
brunoguerios, line 67: The only downside I see from this refactor is that the consumer will have to provide sender/recipient info even if they want to perform a query.
FE asked me to refactor addLiquidityNested and remove `accountAddress` from the query input, so it's not a requirement.
I know they can come around this by providing ADDRESS_ZERO, but I think it's nice to be explicit on the requirements for each method.
Following that, having sender/recipient on the buildCall only would make more sense.
What do you think?

test/entities/swapV3.integration.test.ts
```diff
@@ -0,0 +1,335 @@
...
+            {
+                address: TOKENS[chainId].BAL.address,
+                decimals: TOKENS[chainId].BAL.decimals,
+            },
+            {
+                address: TOKENS[chainId].WETH.address,
+                decimals: TOKENS[chainId].WETH.decimals,
+            },
+        ],
+        pools: ['0xB7FdEa33364Da24d6ad01C98EFAb7b539B917A83'],
```
brunoguerios, line 60: You should be able to use `POOLS[chain.id].MOCK_WEIGHTED_POOL.id`

brunoguerios (commented):
Overall looks really good 👏 
I left a few comments to discuss - as soon as we reach a conclusion on those we should be good to go 👍

src/entities/swap/index.ts
johngrantuk: Ok, yeah I think this sounds good. In this case we will expect that the consumer always uses WETH for Paths and has no concept of using ZERO address?

examples/swap.ts
johngrantuk: In that case we will should use two different buildCall input types - One for V2 that includes sender/recipient and one for V3 that doesn't. Does that sounds ok? Its a pity we have to force the user to decide but I can't see a way around it.

src/entities/swap/index.ts
brunoguerios: Yeap. Zero address is an implicit requirement that we want to avoid. Having an explicit flag should be a better experience.

examples/swap.ts
brunoguerios: True 😕 
I think that's ok though 👍 
They will need to provide either when building the swap class or when building the call anyway 🤷

src/entities/swap/index.ts
johngrantuk: I refactored this to move wethIsEth to buildCall. As discussed with Daniel, when a user wants to swap using native asset they will query the API using the wrapped token. They can then decide to user wrapped/native by setting the param in buildCall.

test/entities/swapV2.integration.test.ts
```diff
@@ -0,0 +1,210 @@
...
+            expect(updated.token).to.deep.eq(balToken);
+            expect(updated.amount).to.eq(60635225778147n);
+        });
+    });
+    describe('swap should be executed correcly', () => {
+        describe('wethIsEth: false', () => {
+            const swapParams = {
+                chainId,
+                paths: [pathBalWeth],
+                wethIsEth: false,
```
brunoguerios, line 122: Swap doesn't expect wethIsEth as input, right? Is this actually being applied somewhere?

test/entities/swapV2.integration.test.ts
```diff
@@ -0,0 +1,210 @@
...
+            });
+            test('GivenOut', async () => {
+                const swap = new Swap({
+                    ...swapParams,
+                    swapKind: SwapKind.GivenOut,
+                });
+                await assertSwapExactOut(client, rpcUrl, chainId, swap, false);
+            });
+        });
+        describe('wethIsEth: true', () => {
```
brunoguerios, line 139: It took me a while to identify that wethIsEth is set through an input at assertSwapExactIn 😅 
And only then I realized that `doSwap` is actually happening inside `assertSwap`.
I'd say that it's ok to leave as is for now, but it would be nice to have a small low priority task to refactor this later on.
Let me know and I'll create one so we don't forget to get back to it.

test/entities/swapV3.integration.test.ts
```diff
@@ -0,0 +1,209 @@
...
+            [WETH.slot as number, BAL.slot as number],
+            [parseEther('100'), parseEther('100')],
+            undefined,
+            balancerVersion,
+        );
+    });
+
+    describe('query method should return correct updated', () => {
+        test('GivenIn', async () => {
+            const swap = new SwapV3({
```
brunoguerios, line 82: Should we instantiate from Swap class instead?
Then we make sure the whole flow is working as expected and avoid the need to import directly from SwapV3?
Or the goal is to actually isolate things while testing?

brunoguerios (approved):
I added a few small comments, but looks good to me otherwise ✅

src/entities/swap/swapV2/index.ts
```diff
@@ -215,31 +201,53 @@ export class SwapV2 implements SwapBase {
...
-    buildCall(swapCall: SwapCallBuild): SwapBuildOutputBase {
-        const limits = this.limits(swapCall.limitAmount);
-        return {
-            to: this.to(),
-            callData: this.callData(
+    buildCall(swapCall: SwapCallBuildV2): SwapBuildOutputBase {
+        const funds = {
+            sender: swapCall.sender,
+            recipient: swapCall.recipient,
+            fromInternalBalance: false, // Set default to false as not supported in V3 and keeps interface simple
```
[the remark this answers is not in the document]
johngrantuk: I thought it was already defaulting to false in main branch so I assumed it had already been decided. I'd say its ok to remove it for now as there seems to be no use of it and it keeps the interface/choice simpler for the user.

test/entities/swapV2.integration.test.ts
johngrantuk: It doesn't. I must have missed this during the refactor. Removed it now.

test/entities/swapV2.integration.test.ts
johngrantuk: Did a small refactor that should make wethIsEth a bit clearer.

test/entities/swapV3.integration.test.ts
johngrantuk: I think at some point I thought it should isolate SwapV3 but as its an integrations test its probably best to test whole flow. Changed here and for V2 test.

brunoguerios (approved):

PR #252, opened 2024-02-16
feat: Add single swap price impact to Swap class.

johngrantuk:
Updated Swap class to use the singleSwap PriceImpact method (this will only work for V2 ATM).
Updated swap example for BAL>WETH swap should pick a single swap through 80/20 pool. Do you think the PI result looks ok (can compare to UI) @brunoguerios?


brunoguerios:
Closing this in favor of #270

PR #262, opened 2024-02-21
V3 Initialize Pool

lgahdl:
Adds V3's Initialize Pool feature


src/data/providers/initPoolDataProvider.ts
```diff
@@ -105,9 +105,39 @@ export class InitPoolDataProvider {
...
+            return {
+                id: poolAddress,
+                address: poolAddress.toLowerCase() as Address,
+                type: poolType,
+                tokens: poolTokens,
+                balancerVersion: 3,
+            };
+        } catch (e) {
+            console.warn(e);
+            throw new Error('Invalid pool address');
```
brunoguerios, line 140: I think this may be a bit misleading, since the error may also come from `getTokenDecimals`.
Ideally we should be able to remove the try/catch and rely on the errors coming from the contracts, but if they are not clear enough, maybe having one try/catch for each point of failure with a proper error message for each would make more sense.
What do you think?
Ps: this applies for the v2 implementation above
Ps2: getTokenDecimals could have an inner try/catch instead of having an external one here

src/entities/initPool/initPoolV3.ts
```diff
@@ -1,9 +1,48 @@
...
+
+        const call = encodeFunctionData({
+            abi: balancerRouterAbi,
+            functionName: 'initialize',
+            args,
+        });
+
+        const value = input.amountsIn.find(
+            (a) => a.address === ZERO_ADDRESS,
+        )?.rawAmount;
```
brunoguerios, line 31: This won't work for v3 as it does not handle the zero_address as native asset. You'll have to check for the wethIsEth flag (or equivalent)

src/entities/initPool/initPoolV3.ts
```diff
@@ -1,9 +1,48 @@
...
+
+        return {
+            call,
+            to: BALANCER_ROUTER[input.chainId] as Address,
+            value: value === undefined ? 0n : value,
```
brunoguerios, line 36: I feel like having a private `value()` method that encapsulates all the `value` logic is nicer to read than having logic split between here and the variable declaration.
What do you think?

src/entities/inputValidator/inputValidator.ts
```diff
@@ -28,7 +28,7 @@ export class InputValidator {
     }
 
     validateAddLiquidity(
-        addLiquidityInput: AddLiquidityInput | InitPoolInput,
+        addLiquidityInput: AddLiquidityInput | InitPoolInputV2,
```
brunoguerios, line 31: I see a few errors within `validateTokensAddLiquidity` that are caused by having AddLiquidity and InitPool sharing the validator.
How about splitting them? You could also add a validator to the V3 input, which I think is missing.

test/v3/initPool.integration.test.ts
```diff
@@ -0,0 +1,132 @@
...
+    walletActions,
+    parseEther,
+    zeroAddress,
+    parseUnits,
+} from 'viem';
+
+const { rpcUrl } = await startFork(ANVIL_NETWORKS.SEPOLIA);
+const chainId = ChainId.SEPOLIA;
+
+describe('Add Liquidity Init - Weighted Pool', async () => {
```
brunoguerios, line 35: You should be able to remove Add Liquidity from the test description.

test/v3/initPool.integration.test.ts
```diff
@@ -0,0 +1,132 @@
...
+            initPoolTxInput.client,
+            initPoolTxInput.testAddress,
+            [...poolState.tokens.map((t) => t.address)],
+            [3, 1],
+            [...poolState.tokens.map((t) => parseUnits('100', t.decimals))],
+            undefined,
+            3,
+        );
+    });
+    test('Add Liquidity Init - Weighted Pool', async () => {
```
brunoguerios, line 123: Same here

brunoguerios (commented):
I added a few suggestions. Let me know what you think 😉

src/entities/inputValidator/inputValidator.ts
lgahdl: Good call! Indeed, a validator only for the initPool functionality is more appropriate, done!

src/entities/initPool/initPoolV3.ts
```diff
@@ -45,4 +43,15 @@ export class InitPoolV3 implements InitPoolBase {
             exactAmountsIn: getAmounts(tokens, input.amountsIn),
         };
     }
+
+    private value(input: InitPoolInputV3) {
+        return input.wethIsEth
+            ? (input.amountsIn.find((a) =>
+                  isSameAddress(
+                      a.address,
+                      NATIVE_ASSETS[input.chainId].address,
```
brunoguerios, line 52: You'll have to check for `wrapped`, since this is the input that the user has to provide.
```suggestion
                      NATIVE_ASSETS[input.chainId].wrapped,
```

src/entities/inputValidator/weighted/inputValidatorWeighted.ts
```diff
@@ -60,4 +72,24 @@ export class InputValidatorWeighted implements InputValidatorBase {
...
+        initPoolInput: InitPoolInputV3,
+        poolState: PoolState,
+    ) {
+        if (!initPoolInput.wethIsEth) {
+            return;
+        }
+        const poolContainsWrappedNativeAsset = !!poolState.tokens.find((a) =>
+            isSameAddress(
+                a.address,
+                NATIVE_ASSETS[initPoolInput.chainId].address,
```
brunoguerios, line 86: Same here with `wrapped`:
```suggestion
                NATIVE_ASSETS[initPoolInput.chainId].wrapped,
```

src/utils/tokens.ts
```diff
@@ -1,27 +1,71 @@
-import { getContract } from 'viem';
+import { Address, getContract } from 'viem';
+import { vaultV2Abi, vaultV3Abi } from '../abi';
+import { VAULT, VAULT_V3 } from './constants';
 
 export async function getTokenDecimals(tokenAddress, client) {
```
brunoguerios, line 5: How about you add types to these inputs?

src/utils/tokens.ts
```diff
@@ -1,27 +1,71 @@
...
+        return decimals;
+    } catch (e) {
+        console.warn(e);
+        throw new Error(
+            `Error: Unable to get Token Decimals from token: ${tokenAddress}`,
+        );
+    }
+}
+
+export async function getPoolTokensV2(poolId, client) {
```
brunoguerios, line 38: It's much better with these helpers 👏

src/utils/tokens.ts
```diff
@@ -1,27 +1,71 @@
...
+        return poolTokensFromVault;
+    } catch (e) {
+        console.warn(e);
+        throw new Error(
+            `Error: Unable to get pool tokens using this pool id: ${poolId}`,
+        );
+    }
+}
+
+export async function getPoolTokensV3(poolAddress, client) {
```
brunoguerios, line 56: Same here about the input types

brunoguerios (commented):
LGTM 👍 
I'll leave the approval for John, in case he wants to take a final look after the most recent changes.

brunoguerios (approved):

PR #263, opened 2024-02-22
V3 BatchSwaps

johngrantuk:
Biggest change is how we create call for V3 batchSwaps. Each independent path in `swapExactIn/Out` needs to have its `minAmountOut`/`maxAmountIn` set correctly using the user defined slippage (compared to V2 where we just updated a single limit index). Discussed with Daniel/Juani and we belive the Router swap queries (not curently implemented) will be able to return the updated path amounts. These can then be used with user definied slippage to update the relevant paths during call build.


src/entities/swap/swapV3/index.ts
```diff
@@ -164,17 +238,152 @@ export class SwapV3 implements SwapBase {
...
+     * @param pathLimits individual path minAmountOut/maxAmountIn depending on SwapKind
+     * @param deadline
+     * @param wethIsEth
+     * @returns
+     */
+    private callDataBatchSwap(
+        limitAmount: bigint,
+        pathLimits: bigint[],
+        deadline: bigint,
+        wethIsEth: boolean,
```
brunoguerios, line 327: How about we set default values for these? 
- Deadline -> MAX
- wethIsEth -> false

I'd expect very few users to go without these default values, so we might be providing a better experience by filling out the value for them.

brunoguerios (commented):
Swap code is getting quite complex 👀  
Makes sense given that we have to support single and batch swaps from 2 vault versions 😅 
Overall I think looks good 👍 
Will do a final review once Router is updated and we're able to add proper integration tests.

src/entities/swap/swapV3/index.ts
johngrantuk: I wasn't really sure about defaults. I was leaning more to making them explicit requirements so the user understands what is actually happening. Not super tied to the idea though so happy to go with what you prefer.

src/entities/swap/types.ts
```diff
@@ -48,35 +49,48 @@ export interface SwapBase {
...
+    ): Promise<ExpectedExactIn | ExpectedExactOut>;
     queryCallData(): string;
     buildCall(swapCall: SwapCallBuild): SwapBuildOutputBase;
 }
 
-export type SwapCallExactInBase = {
-    deadline: bigint;
-    slippage: Slippage;
-    wethIsEth: boolean;
+export type ExpectedBase = {
```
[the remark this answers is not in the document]
johngrantuk: They should definitely all use the same naming conventions and I _think_ I agree about `QueryOutput`. Have refactored 👍.

src/entities/swap/swapV3/index.ts
johngrantuk: ✅

src/entities/swap/swaps/v3/index.ts
```diff
@@ -0,0 +1,525 @@
...
+            swapKind: SwapKind.GivenOut,
+            expectedAmountIn: TokenAmount.fromRawAmount(
+                this.inputAmount.token,
+                result[2][0],
+            ),
+            pathAmounts: result[0] as bigint[],
+        };
+    }
+
+    public queryCallData(): string {
```
brunoguerios, line 217: This method feels a bit redundant with `query`. Can we build query data and then reuse it within query to avoid code duplication? Or is different in any way?

brunoguerios: And maybe rename to `buildQuery` to be consistent with `buildCall`?

brunoguerios (changes requested):
Overall implementation looks really nice. 👏 
I added some small comments for us to discuss.

Note: if you get the branch up to date with main, it has extended timeouts on the failing tests, so they don't break CI anymore.

test/anvil/anvil-global-setup.ts
```diff
@@ -102,6 +102,23 @@ export async function stopAnvilForks() {
     runningForks = {};
 }
 
+// Stop a specific anvil fork
+export async function stopAnvilFork(
```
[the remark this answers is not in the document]
johngrantuk: It was only really an issue here because the previous tests were removing enough liquidity from pools that made future swap paths impractical.
I do think there is likely a better solution to what I found. Originally I tried [reset](https://viem.sh/docs/actions/test/reset#reset) which I thought would be ideal but it really slows down the tests. Probably worth making an issue to investigate/understand this better as we should really try to optimise for speed and provider calls if possible?

src/entities/swap/swaps/v3/index.ts
johngrantuk: `queryCallData` is a function from the original Swaps implementation. I originally thought it was a bit redundant too but I think the goal is to provide the encoded data so a user can use it to call efficiently whenever and however they want without have to re-encode, etc.
The only way I can see to use the call data in `query` is to use [call](https://viem.sh/docs/actions/public/call) but that returns encoded data which we would then need to decode. `simulate` seems like the better choice here.
I've added a helper, `getSwapsWithDefaultLimits()` that can be used in both to reuse as much common code as possible.

src/entities/swap/types.ts
```diff
@@ -1,14 +1,26 @@
...
+type SwapCallBuildInputBase = {
+    deadline?: bigint;
+    slippage: Slippage;
+    wethIsEth?: boolean;
+    queryOutput: ExactInQueryOutput | ExactOutQueryOutput;
+};
+
+export type SwapCallBuildInput =
+    | SwapCallBuildInputBase
+    | (SwapCallBuildInputBase & SwapCallBuildInputBaseV2);
```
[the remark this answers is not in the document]
johngrantuk: Yep, that was actually annoying me tbh! Nice suggestion about limits. I've refactored quite a bit, there's some code repitition but I think its a good bit more intuitive and easier to follow but let me know what you think?

I'm not sure what the benefits would be from making the main `Swap` object implement the `SwapBase`. It would be easier to do after this refactor but would mostly be adding redundant properties/functions that I think would clutter unnecessarily. Let me know if you think thats the wrong conclusion though.

test/anvil/anvil-global-setup.ts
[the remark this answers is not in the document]
brunoguerios: I created #291 to keep track of this ✅

src/entities/swap/swaps/v3/index.ts
brunoguerios: Got it 👍 
I agree that being able to use simulate is better than reusing a bit more code.

src/entities/swap/types.ts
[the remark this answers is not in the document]
brunoguerios: One possible solution would be to break SwapBase into SwapBase (methods) and SwapState (attributes), and have the external Swap implement only SwapBase.
But I'd say the current code is is no longer confusing, so we should be good to go with it 👍

brunoguerios (approved):
LGTM! 👍

PR #266, opened 2024-02-23
Moving addLiquidity, addLiquidityNested, removeLiquidity and swap exa…

lgahdl:
…mples to v2 folder;

Changing README.md file and adding structure;


examples/v2/addLiquidity.ts
johngrantuk, line 1: The goal is to make these version agnostic so I don't think we should split them into v2/v3.

lgahdl: True! Sorry, my bad

brunoguerios:
Closing this in favor of #293

PR #270, opened 2024-02-29
Add price impact for swaps and batchSwaps

### LINKED ISSUE #264: Price Impact - Swaps
Previous solution for calculating price impact on swaps and batchSwaps relied on the SOR, which is no longer available within b-sdk.

The goal of this task is to provide a new solution for that.

A suggestion on how to approach it using the ABA method would be:
1. Query the SOR to build a swap path between A and B
2. Build a Swap object with SwapInput from the API
3. Query its return amount => B
4. Build a "reverse" Swap with the output (could also be the same path, but givenOut with B amount)
5. Query its return amount => finalA
6. Calculate price impact between A and finalA

brunoguerios:
Close #264

### DIFF
--- a/src/entities/priceImpact/index.ts
+++ b/src/entities/priceImpact/index.ts
@@ -18,7 +18,9 @@ import { TokenAmount } from '../tokenAmount';
 import { PoolState } from '../types';
 import { getSortedTokens } from '../utils';
 import { SingleSwap, SwapKind } from '../../types';
-import { SingleSwapInput, doSingleSwapQuery } from '../utils/doSingleSwapQuery';
+import { doSingleSwapQuery } from '../utils/doSingleSwapQuery';
+import { Swap, SwapInput } from '../swap';
+import { Token } from '../token';
 
 export class PriceImpact {
     static addLiquiditySingleToken = async (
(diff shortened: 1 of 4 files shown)


src/entities/swap/index.ts
```diff
@@ -59,10 +58,6 @@ export class Swap {
         return this.swap.queryCallData();
     }
 
-    public get priceImpact(): PriceImpactAmount {
```
johngrantuk, line 62: 👏

test/v2/priceImpact/priceImpact.integration.test.ts
johngrantuk, line 1: seeing a couple of type errors in this file.

src/entities/priceImpact/index.ts
```diff
@@ -382,52 +384,85 @@ export class PriceImpact {
...
+            );
+        } else {
+            // givenOut case
+            // Note: in the given out case, query output is greater than input, so we
+            // reverse the order of initialA and finalA for ABA to work as expected
+            initialA = TokenAmount.fromRawAmount(
+                tokenA,
+                reverseSwapQueryOutput.expectedAmountIn.amount,
+            );
+            finalA = TokenAmount.fromRawAmount(
```
johngrantuk, line 457: I think you can use `swap.outputAmount` here instead of mapping?

src/entities/priceImpact/index.ts
```diff
@@ -382,52 +384,85 @@ export class PriceImpact {
...
+        const tokenA = new Token(
+            swapInput.chainId,
+            swapInput.paths[0].tokens[tokenAIndex].address,
+            swapInput.paths[0].tokens[tokenAIndex].decimals,
         );
+        let initialA: TokenAmount;
+        let finalA: TokenAmount;
+        if ('expectedAmountOut' in reverseSwapQueryOutput) {
+            // givenIn case
+            initialA = TokenAmount.fromRawAmount(
```
johngrantuk, line 439: I think you can use swap.inputAmount here instead of mapping?

johngrantuk (approved):
Couple of small comments but otherwise looks good. I think its def nicer having the separate Price Impact class.

test/v2/priceImpact/priceImpact.integration.test.ts
brunoguerios: fixed ✅

src/entities/priceImpact/index.ts
brunoguerios: ✅

PR #276, opened 2024-03-06
Refactor useNativeAssetAsWrappedAmountIn into wethIsEth

### LINKED ISSUE #227: Refactor useNativeAssetAsWrappedAmountIn into wethIsEth
There are a few subtasks on this: 

1. rename `useNativeAssetAsWrappedAmountIn` to `wethIsEth`
2. rename `toNativeAsset` to `wethIsEth`
3. move `wethIsEth` from AddLiquidity `query` to `buildCall` interface
4. move `wethIsEth` from RemoveLiquidity `query` to `buildCall` interface

Note: v3 Router interface for Add/Remove Liquidity does not expose a `wethIsEth` input for query operations, but only for the actual transactions, so items 3 and 4 aim to make v2 interface follow the same standard. The assumption is that using ETH or WETH should not affect the query result.

brunoguerios:
Closes #227


test/v2/removeLiquidity/composableStable.recovery.integration.test.ts
```diff
@@ -98,6 +98,7 @@ describe('composable stable remove liquidity test', () => {
                 txInput.slippage,
             );
         });
+        // TODO: with native asset
```
johngrantuk, line 101: Worth adding now?

test/v2/removeLiquidity/weighted.recovery.integration.test.ts
```diff
@@ -103,6 +103,7 @@ describe('weighted remove liquidity recovery test', () => {
                 txInput.slippage,
             );
         });
+        // TODO: with native asset
```
johngrantuk, line 106: Now?

johngrantuk (commented):
Definitely seems like better/easier to follow terminology. Swaps are still using `wethIsEth` should that also be taken care of in this PR to maintain consistency?

test/v2/removeLiquidity/composableStable.recovery.integration.test.ts
brunoguerios: I'd say yes - will add it here 👍

brunoguerios:
Nice! 
Yeap, I split initially to reduce complexity of each commit and forgot about swaps 😅
It makes total sense to update that in this PR as well - will do that now 👍

brunoguerios:
After discussing with the team, Daniel suggested we kept `wethIsEth` for consistency with smart contracts and also across all interfaces (add/remove/swap).
Current changes already take that into account.

src/entities/addLiquidity/types.ts
johngrantuk, line 1: Not sure if its already covered elsewhere but should we take this opportunity to refactor `Add/RemoveLiquidityQueryOutput` to also return chainId so it doesn't have to be passed to build?

johngrantuk (approved):

src/entities/addLiquidity/types.ts
brunoguerios: done ✅

PR #277, opened 2024-03-06
Move accountAddress from query to buildCall input

### LINKED ISSUE #209: Remove the need for accountAddress when querying for Add/Remove Liquidity Nested
Request from @agualis:

> Ideally, it would be good to allow the users to check the “quote” before connecting
> If you are a new user and you are simply exploring balancer is good to have this ability

To achieve that we could to remove `accountAddress` from `AddLiquidityNestedInput` and fill in with `ZERO_ADDRESS`.
Then we'd need to replace that with sender/recipient provided on buildCall.

brunoguerios:
Closes #209


lgahdl (commented):
Looks good, just need to refactor the src/entities/price-impact/index.ts file.
Note: The InitPoolInputV2 also uses sender/recipient params, should it be changed to accountAddress?

src/entities/addLiquidityNested/doAddLiquidityNestedQuery.ts
```diff
@@ -20,7 +19,7 @@ export const doAddLiquidityNestedQuery = async (
     });
 
     const { data } = await client.call({
-        account: accountAddress,
+        account: ZERO_ADDRESS,
```
johngrantuk, line 22: I think account is actually optional so perhaps don't need it at all?

src/entities/removeLiquidityNested/doRemoveLiquidityNestedQuery.ts
```diff
@@ -21,7 +20,7 @@ export const doRemoveLiquidityNestedQuery = async (
     });
 
     const { data } = await client.call({
-        account: accountAddress,
+        account: ZERO_ADDRESS,
```
johngrantuk, line 23: optional?

examples/addLiquidityNested.ts
```diff
@@ -102,8 +101,7 @@ const addLiquidityNested = async () => {
     const { call, to, value, minBptOut } = addLiquidityNested.buildCall({
         ...queryOutput,
         slippage,
-        sender: accountAddress,
-        recipient: accountAddress,
+        accountAddress,
```
johngrantuk, line 104: Are we loosing the option for users to set sender/receiver and is that ok? If yes should this not be made consistent for `addLiquidity` too?

brunoguerios: Talked to Daniel about this and he's going to check with the SC team on how to approach sender/recipient on v3. There's a chance they will update the interface on their end, so we'll leave as is for now and get back to it later if needed.
Nested liquidity V2 is a special case due to the gas cost reduction code that handles transactions between user and relayer, so we'll keep a single `accountAddress` as sender/recipient

johngrantuk (approved):

PR #278, opened 2024-03-06
Refactor sender and recipient to exist on v2 only

### LINKED ISSUE #271: Liquidity buildCall sender/recipient is redundant for V3
Sender/recipient concept no longer exists for V3. 
We still allow it to be passed as a param in add/remove liquidity - this could cause issues as it won't perform as expected for V3 because it will always be send/received from sender.
I attempted to handle this in Swaps by having different input types depending on version for the buildCall function. Then I added some checks for correct input/version [here](https://github.com/balancer/b-sdk/blob/c247595215a6ddad56215ce068b7cabab9dda14e/src/entities/swap/index.ts). Maybe there's a better approach and open to suggestions but I think we should definitely do something for add/remove too.

brunoguerios:
Closes #271


src/entities/addLiquidity/index.ts
```diff
@@ -33,6 +33,13 @@ export class AddLiquidity implements AddLiquidityBase {
     }
 
     buildCall(input: AddLiquidityCall): AddLiquidityBuildOutput {
+        // TODO: refactor validators to take v3 into account
```
johngrantuk, line 36: Is this being tracked in a new issue?

src/entities/addLiquidity/types.ts
```diff
@@ -48,28 +48,32 @@ export type AddLiquidityBaseQueryOutput = {
...
 };
 
-export type AddLiquidityWeightedQueryOutput = AddLiquidityBaseQueryOutput;
-
 export type AddLiquidityComposableStableQueryOutput =
     AddLiquidityBaseQueryOutput & {
         bptIndex: number;
     };
 
+export type AddLiquidityWeightedQueryOutput = AddLiquidityBaseQueryOutput;
```
johngrantuk, line 56: This feels like it leads to things being quite strange/tricky to follow. For example, immediately below `AddLiquidityQueryOutput` is making a Union of `AddLiquidityWeightedQueryOutput` & `AddLiquidityBaseQueryOutput` which are both the same.

johngrantuk (changes requested):
Few initial comments and looks like a build issue too.

src/entities/addLiquidity/index.ts
brunoguerios: I just created #280 to keep track of this ✅

src/entities/addLiquidity/types.ts
brunoguerios: Yeah, I wasn't sure about all this either. There are a couple choices made here:
1. Keeping `AddLiquidityWeightedQueryOutput` alias instead of using `AddLiquidityBaseQueryOutput` for weighted pools directly -> more explicit output name from a user perspective
2. Adding `AddLiquidityBaseQueryOutput` to the union - that's for v3 usage actually, because it doesn't have weighted/csp variations

I could remove the alias and keep it simple - does that sound a better approach? It's not as explicit for the weighted case, but it should be ok 🤷‍♂️

src/entities/addLiquidity/types.ts
```diff
@@ -97,6 +82,7 @@ export type AddLiquidityConfig = {
     customAddLiquidityTypes: Record<string, AddLiquidityBase>;
 };
 
+// type consumed by FE team, so should be exported here
```
johngrantuk, line 85: Do we know what they use it for? Is it SDK related?

johngrantuk (approved):

src/entities/addLiquidity/types.ts
brunoguerios: Hey @groninge01, this was a request from you, right?! Can you please help us remind why that's needed?

brunoguerios: From groninge:
> i think it was because i'd like to batch a joinpool with staking just like we do in our current beets frontend, we encode all data in the frontend and use the type from the 'old/current' sdk
so i asked if the new sdk would do the encoding for us and you said that wasn't a priority then, but instead you could add the type again so we can do the encoding in the frontend (again)

johngrantuk: No big deal but seems like if most of the functionality is already in the FE then the type could also sit locally. Just feels funny sitting here.

brunoguerios: Yeah, I see your point. And I agree it's not a big deal, so if FE team prefer it this way, I'm ok on leaving it there.
I added a more descriptive comment so we don't forget the reason in a few weeks/months again 😅

PR #282, opened 2024-03-08
Additional tidy

johngrantuk:


brunoguerios:
I get the idea of moving type definition closer to their usage, but I must say I get a bit confused having to navigate between files to understand what is expected for each input/output.
I think we can overcome this with more explicit variable names.

Disregarding naming variations, we currently have this:
- AddLiquidityInput - for query inputs
- AddLiquidityQueryOutput - for query outputs
- AddLiquidityCall - for buildCall inputs
- AddLiquidityBuildOutput - for buildCall outputs

What do you think of something like this:
- AddLiquidityInput - for query inputs (they could stay the same, since they are the main input of the add liquidity flow)
- AddLiquidityQueryOutput - for query outputs
- AddLiquidityBuildCallInput - for buildCall inputs
- AddLiquidityBuildCallOutput - for buildCall outputs

If you think this can get too big with variations such as `AddLiquidityV2ComposableStableBuildCallInput` I’m open to simplifications, as long as we have more consistency and it’s easy to infer what they represent 😅

PR #285, opened 2024-03-14
Refactor queryRemoveLiquidityRecovery into a separate method

### LINKED ISSUE #279: Support - b-sdk recovery exits malfunctioning for Phantom Stable pools (Composable Stable V0)
Pool: https://app.balancer.fi/#/ethereum/pool/0x7b50775383d3d6f0215a8f290f2c9e2eebbeceb20000000000000000000000fe
Impersonated Address:
0x26743984e3357eFC59f2fd6C1aFDC310335a61c9
BptIn:
137222.962994584164061051

Image

brunoguerios:
Closes #279 

Main:
- Move queryRemoveLiquidityRecovery into a separate method

Also:
- Refactor calculateProportionalAmounts to be a bit more generic:
    - Accept bptAmount as input
    - Rename variables so it's not relative to addLiquidity only (e.g. amountsIn/bptOut -> tokenAmounts/bptAmount)

### DIFF
--- a/src/entities/removeLiquidity/index.ts
+++ b/src/entities/removeLiquidity/index.ts
@@ -5,11 +5,15 @@ import {
     RemoveLiquidityConfig,
     RemoveLiquidityInput,
     RemoveLiquidityQueryOutput,
+    RemoveLiquidityRecoveryInput,
 } from './types';
-import { PoolState } from '../types';
+import { PoolState, PoolStateWithBalances } from '../types';
 import { InputValidator } from '../inputValidator/inputValidator';
 import { RemoveLiquidityV2 } from './removeLiquidityV2';
 import { RemoveLiquidityV3 } from './removeLiquidityV3';
+import { calculateProportionalAmounts } from '../utils';
+import { TokenAmount } from '../tokenAmount';
+import { Token } from '../token';
 
 export class RemoveLiquidity implements RemoveLiquidityBase {
     private readonly inputValidator: InputValidator = new InputValidator();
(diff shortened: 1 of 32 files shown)


src/entities/removeLiquidity/index.ts
```diff
@@ -33,6 +34,38 @@ export class RemoveLiquidity implements RemoveLiquidityBase {
         }
     }
 
+    /**
+     * It's not possible to query Remove Liquidity Recovery in the same way as
+     * other remove liquidity kinds, so a separate handler is required for it.
+     * Since it's not affected by fees or anything other than pool balances,
+     * it's possible to calculate amountsOut as proportional amounts.
+     */
+    public queryRemoveLiquidityRecovery(
```
johngrantuk, line 43: Do you think its a bit misleading/confusing calling this `query`? Based off the other methods named this way I would expect it to get an up to date onchain result when called.

brunoguerios: My reasoning for keeping a similar method name is because even though it uses a different approach to "query", it returns the same results as the other ones. And should be used in the same way: query -> buildCall
Not sure if external users would mind how the "query" is performed, as long as it behaves the same way.

I get that it's using api balances instead of on-chain, so it doesn't necessarily behave the same way, but it should be pretty straightforward to fetch on-chain balances instead of relying on api, which I agree would make more sense. Would that be enough? Or do you think we'd still be better off with another method name?

johngrantuk: I think the issue I was imagining was a user making a query then waiting some time before making another query to get an updated result. In this case it wouldn't give any update, even if the state had changed. That is different from how I would expect it to work. 
I think as long as it was changed to always use the latest state when query was called then it doesn't matter how that is done.

brunoguerios: done ✅

PR #293, opened 2024-03-27
Docs examples update

### LINKED ISSUE #247: b-sdk - Update README.md and examples

johngrantuk:
Closes #247 

Updating examples to use consistent setup, helpers for forks, etc. Also re-organised and tidied a bit. This will allow easier re-use in follow up docs/readme updates.


examples/lib/getSlot.ts
```diff
@@ -0,0 +1,30 @@
+import { ChainId } from '@/utils';
+import { TOKENS } from 'test/lib/utils';
+import { Address } from 'viem';
+
+export function getSlot(chainId: ChainId, tokenAddress: Address): number {
```
brunoguerios, line 5: Nice helper 👏 
It would be nice to have a small description of what slot means, since it's such a low level thing, right?
Something like:
> slot refers to the storage slot responsible for token balance within an ERC20 contract. It is used to artificially manipulate account balance before any tests/examples on local forks.

examples/lib/getSlot.ts
```diff
@@ -0,0 +1,30 @@
...
+            `TOKENS does not contain token ${tokenAddress} ${chainId}`,
+        );
+    if (!token.slot)
+        throw new Error(
+            `TOKENS does not contain slot for token ${tokenAddress} ${chainId}`,
+        );
+    return token.slot;
+}
+
+function test() {
```
brunoguerios, line 21: Is this test still necessary?

brunoguerios (approved):
I left some small comments, but overall LGTM 👍

lgahdl (commented):
LGTM!

examples/addLiquidity/addLiquidity.ts
```diff
@@ -18,35 +14,76 @@ import {
...
+    // 80BAL-20WETH
+    const pool = {
+        id: '0x5c6ee304399dbdb9c8ef030ab642b10820db8f56000200000000000000000014',
+        address: '0x5c6Ee304399DBdB9C8Ef030aB642B10820DB8F56' as Address,
+    };
+    const amountsIn = [
+        {
+            rawAmount: parseEther('1'),
+            decimals: 18,
+            address: '0xba100000625a3754423978a60c9317c58a424e3D' as Address,
```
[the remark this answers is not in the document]
johngrantuk: I'm not sure I agree with that one but open to change my mind! Could you elaborate on how it helps? We could agree on some standard for this kind of thing to try and get some consistency across all examples.

examples/createPool/createPool.ts
```diff
@@ -0,0 +1,168 @@
...
+
+    // Builds the create pool call using the SDK
+    const call = await createPoolCall(createPoolInput);
+
+    // Submit the create pool tx and wait for the PoolCreated event to find pool address
+    const poolAddress = await createPool({ client, call, userAccount });
+    console.log('Created Pool Address: ', poolAddress);
+
+    // Build the init pool call using the SDK
+    const initCall = await initPool({
```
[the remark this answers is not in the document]
johngrantuk: Originally it was 2 separate examples but because you can't do init without create it imported the create example and also ran that. It didn't seem super clear what was going on and anyone reading it had to link back and forth which can be a pain if you're not in the IDE. So I thought it was better to just do it all in a single example.
Agree on the name though.

examples/lib/getSlot.ts
johngrantuk: Nope!

PR #294, opened 2024-03-28
refactor: Update readme.

johngrantuk:
Updated readme. Idea is this API type documentation will be kept up to date as part of any future PRs and serves as the source to be copied to docs reference page. Also added some missing info for local setup and links to relevant examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,7 +235,7 @@ const swap = new Swap(swapInput: SwapInput);
 
 | Name               | Type          | Description   |
 | -------------      | ------------- | ------------  |
-| swapInput | [SwapInput]() | Swap input including path information. |
+| swapInput | [SwapInput](./src/entities/swap/types.ts#L8) | Swap input including path information. |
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,112 +16,688 @@ If your platform does not support one of the required features, it is also possi
```
lgahdl, line 16: I think it would be good to have the recommended Node version (v18) somewhere in the Local Setup section.

johngrantuk:
> It's looking great 👏 Did you write all of it down? Or is there some sort of lib or auto-generate script?

I wrote it manually. Couldn't find any easy/reliable tool (but open to suggestions!). Should be easy to maintain gong forward if we make it part of the PR process.

README.md
johngrantuk: Is that normal to have in docs? I usually just look at the package.json. If you can link a good example from another library then I'll copy the same format 🙏

PR #295, opened 2024-04-04
Example fixes

johngrantuk:
Main important fix here is the BalancerApi to handle non-lowercase address/id.

### DIFF
--- a/src/data/providers/balancer-api/modules/pool-state/index.ts
+++ b/src/data/providers/balancer-api/modules/pool-state/index.ts
@@ -186,7 +186,7 @@ export class Pools {
         const { data } = await this.balancerApiClient.fetch({
             query: this.poolStateQuery,
             variables: {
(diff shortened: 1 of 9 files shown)


src/data/providers/balancer-api/modules/pool-state/index.ts
```diff
@@ -186,7 +186,7 @@ export class Pools {
         const { data } = await this.balancerApiClient.fetch({
             query: this.poolStateQuery,
             variables: {
-                id,
+                id: id.toLowerCase(),
```
brunoguerios, line 189: I didn't know that was an issue. I'd say it's ok to have it as a quick fix here, but ideally the API should support non-lowercase inputs as well. I'll create an issue on their repo.

brunoguerios (approved):
LGTM 👍

PR #307, opened 2024-04-25
Default input validator

johngrantuk:
There is currently a dependency on an SDK known PoolType when adding/removing liquidity. This stops the SDK working with new pool types. I've added a `baseInputValidator` that is the default when the pool type is unknown. All known pool types extend this with their specific info which is also an improvement as it makes it easier to follow.


examples/lib/getSlot.ts
```diff
@@ -18,7 +18,7 @@ export function getSlot(chainId: ChainId, tokenAddress: Address): number {
         throw new Error(
             `TOKENS does not contain token ${tokenAddress} ${chainId}`,
         );
-    if (!token.slot)
+    if (token.slot === undefined)
```
brunoguerios, line 21: Nice catch!

brunoguerios (approved):
Much better dev experience and extensibility indeed 👏

examples/lib/getSlot.ts
johngrantuk: Caught it the painful way :)!

PR #308, opened 2024-04-25
304 build function return data callcalldata

johngrantuk:


brunoguerios (approved):
LGTM 👍

examples/createAndInitPool/createAndInitPool.ts
```diff
@@ -145,7 +145,7 @@ const initPool = async ({
 async function createPool({ client, call, userAccount }) {
     const hash = await client.sendTransaction({
         to: call.to,
-        data: call.call,
+        data: call.callData,
```
brunoguerios, line 148: I'd say it's clear enough as `callData`, but looking at the standard here being `data`, should we follow it and use `call.data`?

johngrantuk: I actually thought the same! I think maybe callData is better in our case since it sometimes get returned with other params that aren't transaction specific. Happy to change if you do prefer though but could be done in a follow up PR.

PR #316, opened 2024-05-08
Price impact is erroring out for add/remove liquidity with proportion…

### LINKED ISSUE #315: Price impact is erroring out for add/remove liquidity with proportional amounts
Here are some screenshots of the issue reported by @groninge01:

Image

Image

Requirements:
- For proportional amounts, price impact should return zero instead of error or NaN.
- For amounts that are close enough to being proportional, price impact should also tend to zero instead of error or NaN.

Implementation suggestion:
- add a breaking test for calculating price impact on Add Liquidity Unbalanced operations when inputs are actually proportional
- consider using `calculateProportionalAmounts` + `PoolStateWithBalances` on that specific `blockNumber` to get proportional inputs
- identify and fix the issue

Extra notes:
- from the screenshot, it seems that the error is happening inside `zeroOutDeltas` method. We should be able to spot the edge case and default price impact to zero

mkflow27:
Fixes #315

### DIFF
--- a/test/v2/priceImpact/priceImpact.integration.test.ts
+++ b/test/v2/priceImpact/priceImpact.integration.test.ts
@@ -18,6 +18,7 @@ import {
     NestedPoolState,
     Path,
     PoolState,
+    PoolStateWithBalances,
     PoolType,
     PriceImpact,
     PriceImpactAmount,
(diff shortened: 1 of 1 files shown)


brunoguerios (approved):
Great PR! 👏 
There is only one step left before merging, which is to add the changeset for versioning control.
Let me know if you need any help with that 😉

brunoguerios (approved):
LGTM! 👏

Travis198963 (commented):
Approve

PR #319, opened 2024-05-24
Update b-sdk to work with Permit2 + new testnet deployment

### LINKED ISSUE #298: Adapt SDK to use Permit2
The current solution to interact with the Vault relies on Trusted Routers, which will be replaced with Permit2 signatures.

User will have to approve Permit2 and from then on, approve the Vault to spend tokens based on off-chain signatures.

Main change will be on nested liquidity operations and batchSwaps with remove liquidity steps, because it will require Permit to be signed for each intermediary BPT.

This task might be quite big to tackle at once, so please consider breaking it into smaller subtasks after initial evaluation.

**Pending the release of the new Router that implements this.

brunoguerios:
Closes #298 

This PR covers not only adapting tests to work with new Permit2 flow, but also adapting ABIs, addresses and code to conform with new testnet deployment, such as:
- ABI and addresses:
   - Vault (moved VaultExtension to inside Vault so viem can properly support with type detection)
   - BatchRouter
   - Router
   - WeightedPoolFactory
   - Permit2
- Code:
   - CreatePool
   - approve token helper

I took the opportunity to also:
- Update a minor version of viem - from 2.1.1 to 2.12.1 (no breaking changes)


johngrantuk (approved):
Nice one.

PR #325, opened 2024-06-06
CoW AMM pools - Add support for add/remove liquidity

### LINKED ISSUE #322: SDK: Add support for CoW-AMM pools
Support following operations:
- Add liquidity proportional
- Remove liquidity proportional

brunoguerios:
Closes #322


src/entities/addLiquidity/addLiquidityCowAmm/index.ts
```diff
@@ -0,0 +1,135 @@
...
+
+        const client = createPublicClient({
+            transport: http(input.rpcUrl),
+            chain: CHAINS[input.chainId],
+        });
+
+        const balances: bigint[] = [];
+        for (const token of sortedTokens) {
+            balances.push(
+                await getPoolTokenBalanceCowAmm(
```
johngrantuk, line 49: Instead of doing multiple separate RPC calls can we do a [multicall](https://viem.sh/docs/contract/multicall) (along with totalSupply) to get all the info in a single call?

src/entities/addLiquidity/addLiquidityCowAmm/index.ts
```diff
@@ -0,0 +1,135 @@
...
+        );
+        const tokenInIndex = undefined;
+
+        const output: AddLiquidityBaseQueryOutput = {
+            poolType: poolStateWithBalances.type,
+            poolId: poolStateWithBalances.id,
+            addLiquidityKind: input.kind,
+            bptOut,
+            amountsIn,
+            tokenInIndex,
```
johngrantuk, line 98: We don't have to explicitly return this.

src/entities/addLiquidity/addLiquidityCowAmm/index.ts
```diff
@@ -0,0 +1,135 @@
...
+        const callData = encodeFunctionData({
+            abi: cowAmmPoolAbi,
+            functionName: 'joinPool',
+            args: [amounts.minimumBpt, amounts.maxAmountsInWithoutBpt],
+        });
+
+        return {
+            callData,
+            to: input.poolId,
+            value: getValue(input.amountsIn, !!input.wethIsEth),
```
johngrantuk, line 125: I think the conclusion was native asset is not supported? In that case we should probably check and throw early if it has been incorrectly configured by consumer?

src/entities/addLiquidity/index.ts
```diff
@@ -21,6 +23,13 @@ export class AddLiquidity implements AddLiquidityBase {
     ): Promise<AddLiquidityQueryOutput> {
         this.inputValidator.validateAddLiquidity(input, poolState);
         switch (poolState.vaultVersion) {
+            case 0: {
+                const addLiquidity = new AddLiquidityCowAmm();
```
johngrantuk, line 27: Seems likely there will be a future version of CowAMM pools. Should we already start checking for the new version and throw if not supported?

src/entities/addLiquidity/index.ts
```diff
@@ -33,14 +42,26 @@ export class AddLiquidity implements AddLiquidityBase {
...
-            const addLiquidity = new AddLiquidityV2(this.config);
-            return addLiquidity.buildCall(input);
-        }
-
-        if (input.vaultVersion === 3 && !('sender' in input)) {
-            const addLiquidity = new AddLiquidityV3();
-            return addLiquidity.buildCall(input);
+        switch (input.vaultVersion) {
+            case 0: {
+                // TODO: check if we need to be more restrictive with input type mismatch here
```
johngrantuk, line 47: I guess we don't do any validation for V2/V3 because we assume buildCall input comes from `query` where validation has already been done so probably ok to keep this assumption.

src/entities/addLiquidity/index.ts
```diff
@@ -21,6 +23,13 @@ export class AddLiquidity implements AddLiquidityBase {
     ): Promise<AddLiquidityQueryOutput> {
         this.inputValidator.validateAddLiquidity(input, poolState);
         switch (poolState.vaultVersion) {
+            case 0: {
+                const addLiquidity = new AddLiquidityCowAmm();
+                return addLiquidity.query(
+                    input as AddLiquidityProportionalInput,
```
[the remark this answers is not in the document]
johngrantuk: Ha! Forgot this was done in the inputValidator!

src/entities/removeLiquidity/index.ts
```diff
@@ -23,6 +25,13 @@ export class RemoveLiquidity implements RemoveLiquidityBase {
     ): Promise<RemoveLiquidityQueryOutput> {
         this.inputValidator.validateRemoveLiquidity(input, poolState);
         switch (poolState.vaultVersion) {
+            case 0: {
+                const removeLiquidity = new RemoveLiquidityCowAmm();
```
johngrantuk, line 29: Similar to add - wonder if we should check for version?

src/entities/removeLiquidity/index.ts
```diff
@@ -66,22 +79,29 @@ export class RemoveLiquidity implements RemoveLiquidityBase {
...
-        if (input.vaultVersion === 3 && isV2Input)
-            throw Error('Cannot define sender/recipient in V3');
-        if (input.vaultVersion === 2 && !isV2Input)
-            throw Error('Sender/recipient must be defined in V2');
-
         switch (input.vaultVersion) {
-            case 2: {
-                const removeLiquidity = new RemoveLiquidityV2(this.config);
+            case 0: {
+                // TODO: check if we need to be more restrictive with input type mismatch here
```
johngrantuk, line 85: Similar comment as add.

src/entities/removeLiquidity/removeLiquidityCowAmm/index.ts
```diff
@@ -0,0 +1,141 @@
...
+
+        const client = createPublicClient({
+            transport: http(input.rpcUrl),
+            chain: CHAINS[input.chainId],
+        });
+
+        const balances: bigint[] = [];
+        for (const token of sortedTokens) {
+            balances.push(
+                await getPoolTokenBalanceCowAmm(
```
johngrantuk, line 49: Similar to add about using multicall. Could probably have a common CowAMM helper like `getPoolState` that can be used for add and remove.

test/lib/utils/helper.ts
```diff
@@ -440,3 +440,61 @@ export const forkSetup = async (
...
+ * @param jsonRpcUrl Url with remote node to be forked locally
+ * @param isVyperMapping Whether the storage uses Vyper or Solidity mapping
+ */
+export const forkSetupCowAmm = async (
+    client: Client & PublicActions & TestActions & WalletActions,
+    accountAddress: Address,
+    tokens: Address[],
+    slots: number[] | undefined,
+    balances: bigint[],
+    spender: Address,
```
johngrantuk, line 461: Because this is specifically for CowAMM lets rename spender to pool as it makes it a bit clearer to follow when reading the function alone?

Also some of the doc comments are out of date.

test/v0/addLiquidity.integration.test.ts
johngrantuk, line 1: Instead of `test/v0...` I think it would be better to do something like `test/cowAMM/v0/...` as it makes it clearer and also allows us to handle future versions. (Same for remove)

johngrantuk (changes requested):
Added a few small comments but otherwise LGTM.

src/entities/addLiquidity/index.ts
brunoguerios: This is actually checking for the `vaultVersion` and not the `poolVersion`. 
It's set as zero because that's how the API is setting it for v1 pools, meaning they have no vault.
Newer poolVersions should not affect how this is implemented as long as the contract interface for getting token balance and totalShares doesn't change.

src/entities/removeLiquidity/index.ts
brunoguerios: Similar response - pool version shouldn't affect the code as long as interfaces keep the same, because proportional adds/removes are not dependent on pool model.

test/lib/utils/helper.ts
brunoguerios: I started it with `pool` and later changed to `spender`, but I agree that pool is easier to read 👍

test/v0/addLiquidity.integration.test.ts
brunoguerios: I used `v0` to match the pattern of tests being split by vaultVersion (e.g. v2/v3). The assumption is that we could eventually have other pool types from Balancer v1?
In any case, I agree this is confusing to use v0 in some places and CowAmm on other places. I'll remove this v0 reference and use CowAmm everywhere.
We can always adjust in case we need to support other v1 pools in the future.

src/entities/addLiquidity/addLiquidityCowAmm/index.ts
brunoguerios: This is actually one of the side effects of reusing v2/v3 flow for CowAmm. 
I agree that `tokenInIndex` is indeed not relevant for CowAmm, but AddLiquidityProportional for v2/v3 do return this explicitly and I believe we should keep the same pattern.
This gets highlighted in the test helper that expects this attribute to be set [here](https://github.com/balancer/b-sdk/blob/925a4d8928f4bfa638f9f23b090f6f4d95d73667/test/lib/utils/addLiquidityHelper.ts#L335).
I'd have to add an if clause there and skip this check specifically for CowAmm pools, which I don't think make much sense.
I'd lean towards leaving as is, but let me know if you still prefer to remove it and I'll do it over here 😉

src/entities/addLiquidity/addLiquidityCowAmm/index.ts
brunoguerios: done ✅

src/entities/addLiquidity/addLiquidityCowAmm/index.ts
brunoguerios: done ✅

src/entities/addLiquidity/index.ts
brunoguerios: done ✅

src/entities/removeLiquidity/index.ts
brunoguerios: done ✅

src/entities/removeLiquidity/removeLiquidityCowAmm/index.ts
brunoguerios: done ✅

test/lib/utils/helper.ts
brunoguerios: done ✅

test/v0/addLiquidity.integration.test.ts
brunoguerios: done ✅

src/entities/addLiquidity/addLiquidityCowAmm/index.ts
johngrantuk: Thanks for the info, happy with that.

src/entities/addLiquidity/index.ts
johngrantuk: Sorry, my comment wasn't very clear. I actually meant we should maybe be checking for `poolVersion` too. My reasoning was that there will likely be a CowAMM V2 (which might have a different interface) or a completely different non-vault amm which I assume would be vaultVersion 0 too. Not a big deal but I think throwing an error when its an unexpected version would potentially avoid problems.

johngrantuk: As we talked, lets leave as is.

johngrantuk (approved):
Nice work!

PR #327, opened 2024-06-13
Update pnpm to v9

brunoguerios:
I ended up updating pnpm to v9, so I'm leaving this as a suggestion for everyone to bump their versions as well.

~~PS: I didn't experience any breaking changes, but thought it would be better to have a separate PR for this anyway~~

PS: lint check failed on github actions, so I had to update biome and it introduced a few breaking changes 😬


agualis (approved):

brunoguerios:
Ok, this actually triggered a few extra changes 🤦‍♂️ 
GitHub lint check failed - for some reason this update made it use biome 1.8 instead of 1.5 which was the one we were using (couldn't figure out why 🤔)
Anyway, I decided to bump all dependencies to their latest versions. There were no major updates, so we should be ok.
Even though biome updated a minor version, there were a few breaking changes to their config file 🤷‍♂️ 
I fixed everything and it should be straightforward to update on your end.
Ps: my vscode got a bit lost during these changes and apparently deleting node_modules (using rimraf) and reinstalling them did the trick - so keep that in mind if your auto-save breaks for some reason 😬 
Ps2: why version updates have to be so painful at times?! 🙄

agualis (approved):

debug/sor.ts
johngrantuk, line 1: I think this can be quite useful.

brunoguerios: I removed it because it was using the local SOR which doesn't exist anymore.
Do you mean that we should keep it here and update it to query the API for paths/swaps?

johngrantuk: No, you're right, sorry 🤦‍♂️

Issue #280, opened 2024-03-07
Refactor input validators to take v3 into account


brunoguerios:
Current input validator implementation takes into account only v2. We'll need to think a bit on how to handle v3 as well.

Example: ComposableStable pools on v3 won't contain their own BPT as token.

PR #329, opened 2024-06-14
Add buildCallWithPermit variations

### LINKED ISSUE #323: SDK V3: Permit2 - Add buildCallWithSignatures
Add one extra buildCall function that expects Permit2 signatures as input and returns the final encoded callData.

This applies to the following operations:
- Add Liquidity
- Remove Liquidity
- Swap
- BatchSwap
- Init (?)
- Create (?)

brunoguerios:
Close #323 

Main changes:
- swap with permit2 signatures
- add liquidity with permit2 signatures
- remove liquidity with permit signatures

Also:
- `approveTokens` and `setTokenBalances` were exposed from test helpers, so we have more flexibility when setting up a test
- add/remove/swap test helpers now have an extra `usePermitSignatures` flag
- add/remove/swap tests were split into `permit direct approval` and `permit signatures` which basically replicate tests for those 2 conditions, so we make sure we’re not missing any scenario that needs to be tested
- new input validator for buildCallWithPermit was added

Notes:
- addLiquidity and swap when `wethIsEth = true` is currently not supported by SC (may be fixed in the future, but I suggest we leave it out of scope for this PR)

### DIFF
--- a/src/entities/permit2Helper/index.ts
+++ b/src/entities/permit2Helper/index.ts
@@ -35,8 +35,21 @@ export class Permit2Helper {
         input: AddLiquidityBaseBuildCallInput & {
             client: Client & WalletActions & PublicActions;
             owner: Address;
+            nonces?: number[];
+            expirations?: number[];
         },
     ): Promise<Permit2> {
+        if (input.nonces && input.nonces.length !== input.amountsIn.length) {
+            throw new Error("Nonces length doesn't match amountsIn length");
+        }
+        if (
+            input.expirations &&
+            input.expirations.length !== input.amountsIn.length
+        ) {
+            throw new Error(
+                "Expirations length doesn't match amountsIn length",
+            );
+        }
         const amounts = getAmountsCall(input);
         const spender = BALANCER_ROUTER[input.chainId];
         const details: PermitDetails[] = [];
@@ -48,6 +61,8 @@ export class Permit2Helper {
                     input.owner,
                     spender,
                     amounts.maxAmountsIn[i],
+                    input.expirations ? input.expirations[i] : undefined,
+                    input.nonces ? input.nonces[i] : undefined,
                 ),
             );
         }
@@ -58,6 +73,8 @@ export class Permit2Helper {
         input: SwapBuildCallInputBase & {
             client: Client & WalletActions & PublicActions;
             owner: Address;
+            nonce?: number;
+            expiration?: number;
         },
     ): Promise<Permit2> {
         // get maxAmountIn
@@ -87,6 +104,8 @@ export class Permit2Helper {
                 input.owner,
                 spender,
                 maxAmountIn.amount,
+                input.expiration,
+                input.nonce,
             ),
         ];
 


src/entities/inputValidator/inputValidator.ts
```diff
@@ -91,4 +95,17 @@ export class InputValidator {
         if (chainId in ChainId) return;
         throw new Error(`Unsupported ChainId: ${chainId}`);
     }
+
+    static validateBuildCallWithPermit2(input: {
```
brunoguerios, line 99: This is not following the pattern, but it simplifies the implementation significantly.
To be honest, I don't like the way InputValidator is currently implemented. Seems to be more complicated than it needs to be.
In any case, I think it's ok as it, but let me know what you think

MattPereira: Definitely agree `InputValidator` could be done better and I like the static method pattern you're proposing here. Much easier to read :+1:

src/entities/permit2/domain.ts
```diff
@@ -0,0 +1,20 @@
...
+        chainId,
+        verifyingContract: permit2Address,
+    };
+}
+
+export type PermitData = {
+    domain: TypedDataDomain;
+    types: TypedData;
+    values: any;
+};
```
MattPereira, line 20: Is this type declaration being used?

src/entities/inputValidator/inputValidator.ts
johngrantuk: I don't think I fully appreciate the complexity of following the old pattern here but open to the change if we think it improves things.
If we want to move to a static refactor lets make an issue for it.

src/entities/permit/index.ts
```diff
@@ -0,0 +1,130 @@
...
+ *
+ * @param { Client & WalletActions & PublicActions } client - Wallet client to invoke for signing the permit message
+ */
+const signPermit = async (
+    client: Client & WalletActions & PublicActions,
+    token: Hex,
+    owner: Hex,
+    spender: Hex,
+    amount = MAX_UINT256,
+    nonce?: bigint,
```
johngrantuk, line 66: I think we can tidy this up a bit. Doesn't look like we ever expect nonce to be passed (e.g. in signRemoveLiquidityApproval). 
I think we should:
* move `getNonce` out of `signPermit` and into `signRemoveLiquidityApproval`
then either:
* change `signPermit` nonce to be non-optional or
* change `signRemoveLiquidityApproval` to accept an optional nonce

src/entities/permit/index.ts
johngrantuk, line 1: How about renaming parent folder to `PermitHelper` which keeps with the general standard we use?
Also I think there's some stuff being exported that don't need to be?

src/entities/permit2/allowanceTransfer.ts
```diff
@@ -0,0 +1,193 @@
...
+const PERMIT_BATCH_TYPES = {
+    PermitDetails: PERMIT_DETAILS,
+    PermitBatch: [
+        { name: 'details', type: 'PermitDetails[]' },
+        { name: 'spender', type: 'address' },
+        { name: 'sigDeadline', type: 'uint256' },
+    ],
+} as const;
+
+function isPermit(permit: PermitSingle | Permit2Batch): permit is PermitSingle {
```
johngrantuk, line 67: Never seen this type of thing before: `permit is PermitSingle` seems like its just another way of saying `boolean` or is there a specific reason for it?

src/entities/swap/swaps/v3/index.ts
```diff
@@ -577,4 +609,10 @@ export class SwapV3 implements SwapBase {
         }
         return swaps;
     }
+
+    private to(): Address {
```
johngrantuk, line 613: Do you think this is actually useful? I find it less explicit.

test/lib/utils/helper.ts
johngrantuk, line 1: While we are refactoring this how about also updating the `approveSpender` type functions to first read `hasApprovedToken` and only write the approval if needed?

johngrantuk (changes requested):
So nice to see this all coming together and tests passing!
Few comments/changes requested, also:
* we'll need to add some info to the docs about the various permit nuances. Can you create an issue for this (that includes the relevant TLDR) and make sure its added to SDK project
* In the main comment it says:
"addLiquidity and swap when wethIsEth = true is currently not supported by SC (may be fixed in the future, but I suggest we leave it out of scope for this PR)"
- but it looks like swap tests do have wethIsEth=true. Can you explain a bit more detail about what is/isn't supported.

src/entities/inputValidator/inputValidator.ts
brunoguerios: This buildCallWithPermit input validation is also used by swaps, which don't follow the InputValidator structure. 
The static alternative was an attempt to get around the overhead of that structure and still keep things close together.
There is already an issue about refactoring input validators [here](https://github.com/balancer/b-sdk/issues/280). We could/should expand on that and try to find a new design that is simpler/easier to follow and also a bit more flexible.
It doesn't even need to follow that static pattern if we find another solution that works better considering the whole implementation.
Let me know if you think that's not enough reason to break pattern and I'll adjust over here until we spend proper time on this later on.

src/entities/permit2/domain.ts
brunoguerios: Apparently not 😅 I adapted the implementation from another repo and I forgot to delete it.
Will remove 👍

src/entities/permit/index.ts
brunoguerios: Yeap, that makes sense 👍

src/entities/swap/swaps/v3/index.ts
brunoguerios: Yeah, I changed because I found myself repeating the code of checking for `isBatchSwap` and assigning `to` to the proper address, but I agree that having these things closer to where it's being used is more explicit.
I'll rollback that change 👍

src/entities/permit/index.ts
brunoguerios: A natural extension of this logic is to allow for `nonce` and `deadline` to be provided as input on permit2. The implementation is a bit more complex because user has to provide these configs for each tokenIn. I tried to keep it as flexible as possible. Let me know what you think.

brunoguerios:
I agree that it would be nice to document Permit/Permit2 signatures nuances. I created #369 so we don't forget to do it. Just to be clear, the idea is to add content to readme (including eventual limitations, such as `does not support wethIsEth = true`) and also some examples on how to use it, right?
Ps: I'll wait on a final decision from the SC team to add details about what's supported and what's not - if I understood correctly, everything will be supported in the next testnet deployment.

src/entities/inputValidator/inputValidator.ts
johngrantuk: All good for now I think.

johngrantuk:
> I agree that it would be nice to document Permit/Permit2 signatures nuances. I created #369 so we don't forget to do it. Just to be clear, the idea is to add content to readme (including eventual limitations, such as `does not support wethIsEth = true`) and also some examples on how to use it, right? Ps: I'll wait on a final decision from the SC team to add details about what's supported and what's not - if I understood correctly, everything will be supported in the next testnet deployment.

Replying on the issue directly.

johngrantuk (approved):
Lets release this beast!!

PR #335, opened 2024-06-24
Update to 5th testnet release of v3

### LINKED ISSUE #334: SDK V3: Update to 5th testnet deployment of contracts
### Overview
We need to update the contract addresses and ABIs to be the latest deployments. 

### Resources
- Example PR  :point_right:  https://github.com/balancer/b-sdk/pull/319
- [Latest contract addresses from 5th testnet deployment](https://github.com/balancer/pvt-deployments/blob/d38e57bfd22dc021b1209de871a51d61ea6c550b/tasks/v3-vault-5/output/sepolia.json)
- [Verified contract info on Tenderly](https://dashboard.tenderly.co/balancer/v3/contracts) (grab ABIs here)
- Mock Pool Deployments used for tests
```
  "B-50BAL-50WETH": "0x7CF221Fa36584f59A4F7FD7b946B8571c78E3692",
  "B-50DAI-50USDC": "0xf45f9638460e65173eAc117d9CE569cC0D561bF6",
  "B-50BAL-50DAI": "0xf6C927cAD5Fe9eB259e59ef604FC74532075dD97",
  "B-50WETH-50USD": "0xB1205100512690D5d51AddFf12AE0Ed99BCe80e8"
```

MattPereira:
Closes #334 

### Summary
- Updated contract addresses and ABIs for the Vault, Router, BatchRouter, and WeightedPoolFactory
- Updated contract addresses for the mock pools that are used for testing
- Adjusted for the addition of `poolHooksContract` to pool creation params
- Adjusted for addition of `isBuffer` and hardcoded to false for now. I agree with [Bruno who suggested handling the logic in a future PR](https://balancerecosystem.slack.com/archives/C03KBCGJL93/p1719358904825109?thread_ts=1719358545.066549&cid=C03KBCGJL93)


src/entities/createPool/createPoolV3/weighted/createPoolWeighted.ts
```diff
@@ -49,7 +49,7 @@ export class CreatePoolWeightedV3 implements CreatePoolBase {
         const roleAccounts: PoolRoleAccounts = {
             pauseManager: input.pauseManager,
             swapFeeManager: input.swapFeeManager,
-            poolCreator: input.poolCreator,
+            poolCreator: zeroAddress,
```
MattPereira, line 52: This one was pretty interesting. The SDK was able to somehow complete the tx to create a weighted pool, but when [simulating with exact same calldata on tenderly](https://dashboard.tenderly.co/balancer/v3/simulator/bac78934-f91a-4925-a118-da342cd4ac6a?trace=0.0), the tx reverted because the monorepo's `WeightedPoolFactory` requires `roleAccounts.poolCreator` to be the zero address.

https://github.com/balancer/balancer-v3-monorepo/blob/98b9efd5d1446ec6e80bb3893374d51601255375/pkg/pool-weighted/contracts/WeightedPoolFactory.sol#L64-L66

test/entities/swapV3.integration.test.ts
```diff
@@ -40,7 +40,7 @@ import { Path } from '@/entities/swap/paths/types';
 const vaultVersion = 3;
 const chainId = ChainId.SEPOLIA;
 // blockNo shouldn't change as checks depend on token balances
-const blockNo = 6001201n;
+const blockNo = 6178561n;
```
MattPereira, line 43: Set to a few blocks after the last mock pool deployment since swaps involve all 4 mock pools

test/anvil/anvil-global-setup.ts
```diff
@@ -52,7 +52,7 @@ export const ANVIL_NETWORKS: Record<NetworksWithFork, NetworkSetup> = {
         rpcEnv: 'SEPOLIA_RPC_URL',
         fallBackRpc: 'https://sepolia.gateway.tenderly.co',
         port: ANVIL_PORTS.SEPOLIA,
-        forkBlockNumber: 6000510n,
+        forkBlockNumber: 6178545n,
```
MattPereira, line 55: Set to a few blocks after the `MOCK_WETH_BAL_POOL` deployment as suggested [here](https://github.com/balancer/b-sdk/issues/334#issuecomment-2188311505)

src/utils/tokens.ts
```diff
@@ -69,7 +69,7 @@ export async function getPoolTokensV3(
     try {
         const chainId = await client.getChainId();
         const vaultV3 = getContract({
-            abi: vaultV3Abi,
+            abi: vaultExtensionV3Abi,
```
MattPereira, line 72: I guess `getPoolTokens` isn't part the Vault ABI anymore?

On scaffold balancer, I've been using the VaultExtension ABI instead of the Vault ABI everywhere. I tried pasting the VaultExtension ABI as the `vaultV3.ts` and the tests all seem to run fine. Would that be prefferrable instead of adding the VaultExtention ABI to the SDK?

test/entities/swapV3.test.ts
```diff
@@ -76,14 +77,17 @@ describe('SwapV3', () => {
...
                             },
                             {
                                 pool: path3hops.pools[1],
                                 tokenOut: path3hops.tokens[2].address,
+                                isBuffer: false,
                             },
                             {
                                 pool: path3hops.pools[2],
                                 tokenOut: path3hops.tokens[3].address,
+                                isBuffer: false,
```
MattPereira, line 90: I don't like veering away from pattern of declaring all path hopping setup in const vars at start of test, but I wasn't sure if its okay or makes sense to add `isBuffer` to the [Path type](https://github.com/balancer/b-sdk/blob/main/src/entities/swap/paths/types.ts#L6)

Also do we need to account for situations where some but not all of the hops will require `isBuffer: true` ?

src/entities/swap/swaps/v3/index.ts
```diff
@@ -532,6 +532,7 @@ export class SwapV3 implements SwapBase {
                             return {
                                 pool: pool,
                                 tokenOut: p.tokens[i + 1].address,
+                                isBuffer: false,
```
MattPereira, line 535: Probably wrong to hardcode these as false, but I'm not sure how to go about writing the logic to toggle `isBuffer` for the swap path steps

src/entities/swap/swaps/v3/index.ts
```diff
@@ -547,6 +548,7 @@ export class SwapV3 implements SwapBase {
...
```
MattPereira, line 551: Don't want to hardcode here either

test/entities/swapV3.integration.test.ts
MattPereira: this block number was too soon :facepalm:

test/entities/swapV3.integration.test.ts
```diff
@@ -344,12 +344,12 @@ describe('SwapV3', () => {
...
                 expect(expected.pathAmounts).to.deep.eq([
-                    2286265697052729n,
-                    831382813409707n,
+                    1827358022063780n,
+                    841746490376824n,
                 ]);
                 expect(expected.expectedAmountIn.token).to.deep.eq(wethToken);
                 expect(expected.expectedAmountIn.amount).to.eq(
-                    3117648510462436n,
+                    2669104512440604n,
```
MattPereira, line 352: Output matches on tenderly
https://dashboard.tenderly.co/balancer/v3/simulator/349d59e8-d990-42c2-9ace-cefd79bfb0a0

test/entities/swapV3.integration.test.ts
```diff
@@ -322,9 +322,9 @@ describe('SwapV3', () => {
                     USDC.decimals,
                 );
                 expect(expected.swapKind).to.eq(SwapKind.GivenIn);
-                expect(expected.pathAmounts).to.deep.eq([90214n, 770252n]);
+                expect(expected.pathAmounts).to.deep.eq([151734n, 779073n]);
                 expect(expected.expectedAmountOut.token).to.deep.eq(usdcToken);
-                expect(expected.expectedAmountOut.amount).to.eq(860466n);
+                expect(expected.expectedAmountOut.amount).to.eq(930807n);
```
MattPereira, line 327: https://dashboard.tenderly.co/balancer/v3/simulator/1e5c70f2-e371-417f-affa-de71ac44abc7

test/entities/swapV3.integration.test.ts
```diff
@@ -118,7 +118,7 @@ describe('SwapV3', () => {
                     WETH.decimals,
                 );
                 expect(expected.expectedAmountOut.token).to.deep.eq(wethToken);
-                expect(expected.expectedAmountOut.amount).to.eq(103365122n);
+                expect(expected.expectedAmountOut.amount).to.eq(123749996n);
```
MattPereira, line 121: https://dashboard.tenderly.co/balancer/v3/simulator/2d2e378e-1ded-4b1e-b4e1-f0fd078575c7

test/entities/swapV3.integration.test.ts
```diff
@@ -133,7 +133,7 @@ describe('SwapV3', () => {
 
                 const balToken = new Token(chainId, BAL.address, BAL.decimals);
                 expect(expected.expectedAmountIn.token).to.deep.eq(balToken);
-                expect(expected.expectedAmountIn.amount).to.eq(96734829986472n);
+                expect(expected.expectedAmountIn.amount).to.eq(80801616032325n);
```
MattPereira, line 136: https://dashboard.tenderly.co/balancer/v3/simulator/79e45b82-7bce-46de-ae99-f91f40e6111c

src/entities/createPool/createPoolV3/weighted/createPoolWeighted.ts
johngrantuk: I wasn't aware of that change! Does indeed seem strange that the SDK tx wasn't reverting.
I think it might be a bit confusing to a user that we allow them to pass a non-zero poolCreator address but it doesn't get used. Maybe we should throw an error if the input isn't zero for this case or change that type?

src/entities/swap/swaps/v3/index.ts
johngrantuk: Yeah, hardcoding is fine for now, we'll have to figure out boosted pools later.

src/utils/tokens.ts
johngrantuk: I can't remember if @brunoguerios did something to combine the abis?
If not I think its maybe best to keep it explicit to match the SCs so using VaultExtension.

test/entities/swapV3.test.ts
johngrantuk: For now this is fine. We'll be revisiting buffer pools later.

johngrantuk (changes requested):
Few small comments/changes but looking good, nice work.

src/entities/createPool/createPoolV3/weighted/createPoolWeighted.ts
brunoguerios: If poolCreator does not exist on WeightedPool (and maybe others), we could/should remove it from `CreatePoolV3BaseInput` and use it only on pool types that actually require it.

src/utils/tokens.ts
brunoguerios: Yeah, I had combined the abis into a single one, but looking again at it, I think it's better if we keep them separated as you did 👍

src/entities/createPool/createPoolV3/weighted/createPoolWeighted.ts
MattPereira: Agreed good thinking @brunoguerios ! :brain:

johngrantuk (approved):

src/abi/balancerBatchRouter.ts
```diff
@@ -1,16 +1,8 @@
...
-                name: 'vault',
-                type: 'address',
-            },
-            {
-                internalType: 'contract IWETH',
-                name: 'weth',
-                type: 'address',
-            },
+            { internalType: 'contract IVault', name: 'vault', type: 'address' },
+            { internalType: 'contract IWETH', name: 'weth', type: 'address' },
```
brunoguerios, line 5: I see that most changes are related to formatting. 🤔 
I believe this is being auto-formatted with Biome, but if that's the case, then why it wasn't already formatted before?
Anyway - let's hope it doesn't change again next time we update the ABI 😅

brunoguerios (approved):
Nice to see how you were able to navigate the SDK, adapt the code, validate results on Tenderly and get the task done 👏 👏 👏

src/abi/balancerBatchRouter.ts
MattPereira: Initially my IDE refused to format because it couldn't find Biome, but it started working after I installed the biomejs extension for vscode

I'll make note of this in the SDK Dev Guide on notion

brunoguerios: On that topic, since most people have Prettier extension installed on vscode, maybe it's also nice to mention that you have to disable it within the workspace for Biome extension to work as expected, otherwise they will conflict.

MattPereira: Agree with sentiment, but I don't think I had to disable my Prettier extension that is setup to format on save

I'm not too familiar with formatters and linters. I'll investigate a bit more

PR #338, opened 2024-06-28
Add aurabal v2 swap support

johngrantuk:
Add support for auraBal<>ETH/WETH/BAL using V2 Relayer.


.vscode/launch.json
johngrantuk, line 1: I don't think this is needed anymore? But let me know otherwise.

src/entities/swap/swaps/v2/auraBalSwaps/joinSwap.ts
```diff
@@ -0,0 +1,133 @@
...
+import { Relayer } from '@/entities/relayer';
+import { auraBalToken, balWethAddress, auraBAL } from './constants';
+import { getJoinData } from './joinPool';
+import { getSwapData } from './swap';
+import { AuraBalSwapQueryOutput } from './auraBalSwaps';
+import { Token } from '@/entities/token';
+import { AuraBalSwapQueryInput } from './parseInputs';
+
+// token[join]8020BPT[swap]auraBAL
+export async function queryJoinSwap(
```
brunoguerios, line 21: It took me a while to understand the scope of these files:
- exitPool
- joinPool
- swap
- joinSwap
- swapExit

I remember these lower level files were under `actions` on the old SDK and I believe that helps navigate things. What do you think about reorganizing to something like this?
- joinSwap
- swapExit
- actions
   - exitPool
   - joinPool
   - swap

src/entities/swap/swaps/v2/auraBalSwaps/joinSwap.ts
```diff
@@ -0,0 +1,133 @@
...
+    // join BAL-WETH 80/20 Pool with joinToken and get 8020BPT in return (to the RELAYER)
+    const { joinPoolData, joinPoolOpRef, value } = getJoinData(
+        joinToken,
+        userAddress,
+        inputAmount,
+        wethIsEth,
+    );
+
+    // Older pools don't have pre-approval so need to add this as a step, approves Vault to spend on RELAYERS behalf
+    const approval = Relayer.encodeApproveVault(balWethAddress, joinPoolOpRef);
```
brunoguerios, line 97: This one took me a while 😅 - I had to go to the old code to understand it.
I didn't know about it and I never realized this was present in the old code 👀 
Not sure how useful that info is from now on, but now I know 😂

brunoguerios (approved):
Oof - that was a long one 😅 

Despite having to know a ton of minor details regarding relayer logic, such as peek, approvals, etc.. to be able to understand the implementation, I feel like it's well done!  

It's a pity that we don't have more time to spend on this right now, because it seems that most of the complexity was already handled and it wouldn't require much more effort to generalize and support any combination of swap/add/remove.

On the other hand, I feel like having to come back to this in a couple months to fix/expand something would be a lot more time consuming 😬

I added a few comments, mostly regarding names/imports that are non critical - otherwise LGTM ✅

examples/swaps/auraBalSwap.ts
```diff
@@ -0,0 +1,139 @@
...
+const auraBalSwap = async ({ rpcUrl, client, userAccount, chainId }) => {
+    // Create AuraBalSwap instance which can be used to query & buildCalls
+    const auraBalSwap = new AuraBalSwap(rpcUrl as string);
+
+    const tokenIn = new Token(
+        1,
+        '0xba100000625a3754423978a60c9317c58a424e3D', // BAL
+        18,
+    );
+    const tokenOut = auraBalToken;
```
MattPereira, line 36: Can you help me better understand context/audience for this example script? I'm curious why import `auraBalToken` from constants for `tokenOut` but show the explicit creation for `tokenIn`. 


Explicit addresses and creation all seem to be shown in examples for v3 SDK docs

test/entities/swaps/v2/auraBalSwaps/auraBal.integration.test.ts
```diff
@@ -0,0 +1,136 @@
...
+import { forkSetup, sendTransactionGetBalances } from 'test/lib/utils';
+import {
+    auraBalToken,
+    BAL,
+} from '@/entities/swap/swaps/v2/auraBalSwaps/constants';
+import { SwapKind } from '@/types';
+
+const chainId = 1;
+const bal = new Token(1, BAL, 18);
+const weth = new Token(1, NATIVE_ASSETS[1].wrapped, 18);
```
MattPereira, line 23:
```suggestion
const chainId = ChainId.MAINNET;
const bal = new Token(chainId, BAL, 18);
const weth = new Token(chainId, NATIVE_ASSETS[1].wrapped, 18);

File: src/entities/swap/swaps/v2/auraBalSwaps/swapExit.ts [hunk 5]
```diff
@@ -0,0 +1,143 @@
...
+    Address,
+    Hex,
+} from 'viem';
+import { TokenAmount } from '../../../../tokenAmount';
+import { BALANCER_RELAYER } from '@/utils';
+import { balancerRelayerAbi } from '@/abi';
+import { Relayer } from '@/entities/relayer';
+import { auraBalToken, balWethAddress, auraBAL } from './constants';
+import { getSwapData } from './swap';
+import { AuraBalSwapQueryOutput } from './auraBalSwaps';
```
MattPereira, line 15:
```suggestion
import { AuraBalSwapQueryOutput } from './types';

File: src/entities/swap/swaps/v2/auraBalSwaps/parseInputs.ts [hunk 6]
```diff
@@ -0,0 +1,46 @@
...
+import { AuraBalSwapKind, SwapQueryInput } from './types';
+import { supportedTokens, auraBalToken } from './constants';
+import { SwapKind } from '@/types';
+import { TokenAmount } from '@/entities/tokenAmount';
+
+export interface AuraBalSwapQueryInput {
+    kind: AuraBalSwapKind;
+    swapToken: Token;
+    inputAmount: TokenAmount;
+}
```
MattPereira, line 12: Might want to move this declaration into `types.ts` next to `AuraBalSwapQueryOutput`

test/entities/swaps/v2/auraBalSwaps/parseInputs.test.ts
```diff
@@ -0,0 +1,81 @@
...
+        expect(async () => {
+            parseInputs({
+                tokenIn: usdc,
+                tokenOut: auraBalToken,
+                kind: SwapKind.GivenIn,
+                swapAmount: TokenAmount.fromHumanAmount(usdc, '1'),
+            });
+        }).rejects.toThrowError('Not A Valid AuraBal Swap');
+    });
+    test('should throw for non-supported input token, FromAuraBal', async () => {
```
MattPereira, line 21:
```suggestion
    test('should throw for non-supported output token, FromAuraBal', async () => {

File: src/entities/swap/swaps/v2/auraBalSwaps/parseInputs.ts [hunk 8]
```diff
@@ -0,0 +1,46 @@
...
+    );
+}
+
+export function parseInputs(input: SwapQueryInput): AuraBalSwapQueryInput {
+    const { tokenIn, tokenOut, swapAmount } = input;
+    if (
+        !isAuraBalSwap(input) ||
+        !tokenIn.isSameAddress(swapAmount.token.address)
+    )
+        throw Error('Not A Valid AuraBal Swap');
```
MattPereira, line 32: Not sure if this matters much, but you might consider modularizing some of the validation checks to throw slightly more specific error messages that hint _why_ the attempted swap input is invalid

src/entities/swap/swaps/v2/auraBalSwaps/auraBalSwaps.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    public isAuraBalSwap(input: SwapQueryInput) {
+        return isAuraBalSwap(input);
+    }
+
+    public async query(input: SwapQueryInput): Promise<AuraBalSwapQueryOutput> {
+        const inputs = parseInputs(input);
+
+        if (inputs.kind === AuraBalSwapKind.ToAuraBal)
+            return queryJoinSwap({ ...inputs, client: this.client });
```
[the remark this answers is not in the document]
johngrantuk: I think its better to keep it as join/exit as it keeps consistency with the SC interfaces that are eventually called. Its only (I think) used internally too so the consumer should never be exposed to it but I think it makes it a bit more explicit and easier to follow when you read the code.

src/entities/swap/swaps/v2/auraBalSwaps/replaceWrapped.ts
```diff
@@ -0,0 +1,13 @@
+import { NATIVE_ASSETS, ZERO_ADDRESS } from '@/utils';
+import { Address } from 'viem';
+
+export function replaceWrapped(tokens: Address[], chainId: number): Address[] {
```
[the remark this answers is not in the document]
johngrantuk: Unfortunately the other one uses `Token` array instead of `Address`. I did think about altering it to handle both but I think its better to keep it more specific.

test/entities/swaps/v2/auraBalSwaps/auraBal.integration.test.ts
```diff
@@ -0,0 +1,136 @@
...
+import {
+    createTestClient,
+    http,
+    publicActions,
+    walletActions,
+    zeroAddress,
+} from 'viem';
+import { Relayer, Slippage, Token, TokenAmount } from '@/entities';
+import { AuraBalSwap } from '@/entities/swap/swaps/v2/auraBalSwaps/auraBalSwaps';
+import { BALANCER_RELAYER, CHAINS, NATIVE_ASSETS } from '@/utils';
```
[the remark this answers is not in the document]
johngrantuk: Was using the example as a check but yeah think its a good idea to do it here too.

src/entities/swap/swaps/v2/auraBalSwaps/auraBalSwaps.ts
```diff
@@ -0,0 +1,100 @@
...
+                input.queryOutput.inputAmount.amount,
+                limitAmount.amount,
+                input.queryOutput.expectedAmountOut.token,
+                input.wethIsEth,
+                input.relayerApprovalSignature,
+            );
+        }
+
+        return {
+            to: BALANCER_RELAYER[1],
```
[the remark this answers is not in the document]
johngrantuk: Good suggestion! The hardcoded `1` was annoying me but for some reason I never thought of using ChainId!

src/entities/swap/swaps/v2/auraBalSwaps/swap.ts
```diff
@@ -0,0 +1,52 @@
+import { Address, encodeFunctionData, Hex } from 'viem';
+import { MAX_UINT256 } from '@/utils';
+import { batchRelayerLibraryAbi } from '@/abi';
+import { Relayer } from '@/entities/relayer';
+import { auraBalStableId } from './constants';
+
+export function getSwapData(
```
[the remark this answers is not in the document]
johngrantuk: Agreed!

src/entities/swap/swaps/v2/auraBalSwaps/joinSwap.ts
johngrantuk: I'll leave it as is for now, maybe if we go back and refactor to be more generalised this would be worthwhile.

examples/swaps/auraBalSwap.ts
johngrantuk: It was mainly to serve as an example to FE team so they can see how to use it. I originally thought using `auraBalToken` was helpful but I think you're right and its better to keep the same standard.

test/entities/swaps/v2/auraBalSwaps/auraBal.integration.test.ts
johngrantuk: Also agree!!

src/entities/swap/swaps/v2/auraBalSwaps/swapExit.ts
johngrantuk: Good catch! Any idea why this doesn't get flagged in CI? Was also the case for joinSwap but I missed both.

MattPereira (approved):
Great work on this one  @johngrantuk :rocket: 

Thanks for thoughtfully responding to all our feedback. It's a pleasure to be working with you!

src/entities/swap/swaps/v2/auraBalSwaps/swapExit.ts
MattPereira: that is interesting. we would expect the lint check to have been failing before huh

I tried reverting the code for this file and running the `pnpm lint` command which is what the `Lint` job from the `Checks` workflow executes during the GHA. Everything still passed even with broken import :thinking:

brunoguerios (approved):
LGTM ✅

PR #348, opened 2024-07-09
feat: Add buffer support to swaps.

johngrantuk:
WIP - Needs updated for deploy6 with query/swap fixes.

Add buffer support. Assumes SOR will add additional field: `isBuffer: boolean[]` as part of `Path` return. Each isBuffer array item at index, `i`, should indicate if corresponding pool in `pools[i]` is a buffer pool. If not returned/undefined we assume there are no buffers in swap path.


test/lib/utils/swapHelpers.ts
```diff
@@ -19,15 +19,46 @@ import {
...
+    rpcUrl,
+    chainId,
+    swap,
+    wethIsEth,
     usePermit2Signatures = false,
-) {
+    outputTest = {
+        isExactOut: true,
+        percentage: 0,
+    },
```
brunoguerios, line 48: I get the idea of differentiating exact match and non-exact match assertions, but I think `isExactOut` flag in an `assertSwapExactIn` check is a bit odd 😅 (same goes for `assertSwapExactOut` with `isExactIn`).
I'd go with simply an optional `percentage`, `withinPercentage` or even `tolerance`. When defined, it specifies the acceptable tolerance and when undefined performs an exact match assertion.
What do you think?

src/abi/index.ts
```diff
@@ -1,6 +1,7 @@
 export * from './authorizer';
 export * from './balancerBatchRouter';
 export * from './balancerQueries';
+export * from './balancerBatchRouter';
```
brunoguerios, line 4: This is already exported above.

brunoguerios (approved):
I left some minor comments on variable naming, but other than that LGTM ✅

test/lib/utils/swapHelpers.ts
johngrantuk: Ahh, yeah, I didn't think of it that way! I think its still useful to make it more explicit about where the percentage is changed so I've kept a flag but changed the naming.

PR #372, opened 2024-07-30
Round down calculateAmountsProportional amounts

brunoguerios:
While comparing `calculateProportionalAmounts` with `queryAddLiquidityProportional` results, we see a small diff of 1 to 2 wei.
By rounding down `calculateProportionalAmounts` results, we should be able to ensure it's always slightly smaller than `queryAddLiquidityProportional` results, which is useful to prevent add liquidity transactions from failing when user is providing all its balance as input.


src/entities/utils/calculateProportionalAmounts.ts
johngrantuk, line 1: Can we add some docs to the `referenceAmount` input param explaining that it should be the token that the user has the lowest balance.

src/entities/utils/calculateProportionalAmounts.ts
```diff
@@ -15,6 +16,7 @@ export function calculateProportionalAmounts(
         tokens: { address: Address; balance: HumanAmount; decimals: number }[];
     },
     referenceAmount: InputAmount,
+    roundDown = true,
```
johngrantuk, line 19: I don't think we should make this optional - if we think this is the best way forward we should just use it and not confuse things with the option.

brunoguerios: I agree the option may be confusing. I added it because, when rounding down, the output may return even your referenceAmount 1 wei less than the input provided. Anyway, the helper is for us to use and it should return what we need. I'll update over here ✅

johngrantuk (approved):

PR #373, opened 2024-08-05
fix: add liquidity example

mkflow27:
This pr fixes the add-liquidity-example. The following changes were made:
- replace the api endpoint. Some V3 pools are not part of the production api endpoint. At a further point in time we need to change it back to the production endpoint
- Reduce add liquidity amount as otherwise the priceImpact calculation (swap queries) will revert with the pool used, due to `revert MaxAmountInRatio`
- Add the chain parameter to the api query
- Added chain name strings the api requires to constants. I initially tried accessing the `Chain.name` from https://github.com/wevm/viem/blob/main/src/types/chain.ts#L49. However, even with uppercasing them they do not perfectly match with what the api expects.

### DIFF
--- a/src/data/providers/balancer-api/modules/pool-state/index.ts
+++ b/src/data/providers/balancer-api/modules/pool-state/index.ts
@@ -2,10 +2,12 @@ import { BalancerApiClient } from '../../client';
 import { PoolState, PoolStateWithBalances } from '../../../../../entities';
 import { mapPoolType } from '../../../../../utils/poolTypeMapper';
 
+import { API_CHAIN_NAMES } from '../../../../../utils/constants';
+
 export class Pools {
     readonly poolStateQuery = `
-    query GetPool($id: String!) {
-      poolGetPool(id:$id) {
+    query poolGetPool($id: String!, $chain: GqlChain!) {
+      poolGetPool(id: $id, chain:$chain) {
         id
         address
         type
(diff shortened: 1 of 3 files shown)


johngrantuk:
@brunoguerios - "Reduce add liquidity amount as otherwise the priceImpact calculation (swap queries) will revert with the pool used, due to revert MaxAmountInRatio" - is that a bug in priceImpact?

src/data/providers/balancer-api/modules/pool-state/index.ts
```diff
@@ -2,10 +2,12 @@ import { BalancerApiClient } from '../../client';
 import { PoolState, PoolStateWithBalances } from '../../../../../entities';
 import { mapPoolType } from '../../../../../utils/poolTypeMapper';
 
+import { API_CHAIN_NAMES } from '../../../../../utils/constants';
+
 export class Pools {
     readonly poolStateQuery = `
-    query GetPool($id: String!) {
-      poolGetPool(id:$id) {
+    query poolGetPool($id: String!, $chain: GqlChain!) {
```
johngrantuk, line 9: I didn't realise the API had changed to require the chain field. Seems nicely handled 👍

johngrantuk (approved):
LGTM and runs locally.

johngrantuk:
@brunoguerios - any further comments before this gets merged?

brunoguerios (approved):
LGTM ✅

PR #380, opened 2024-08-07
Fix outputReference index being incorrectly consumed

brunoguerios:
OutputReference index was being incorrectly passed down to the next call input.


brunoguerios:
For context, outputReference indexes are based on pool tokens `without BPT`. That's why sometimes there's a mismatch on token index and outputReference index. The fix should handle all scenarios correctly from now on.

PR #387, opened 2024-08-13
Fix add liquidity proportional query for cow-amm

brunoguerios:
Context: Add Liquidity Proportional Query for CowAmm pools rely on getting current pool state onChain and then calculating proportional amounts from it.
Existing calculateProportionalAmounts implementation was not matching the SC implementation for CowAmm pools, resulting in a few wei difference between them.
This PR adds a new calculateProportionalAmountsCowAmm that replicates exactly the math done in the SC, resulting in an exact match between their results.

### DIFF
--- a/.changeset/sixty-squids-report.md
+++ b/.changeset/sixty-squids-report.md
@@ -1,0 +1,5 @@
+---
+"@balancer/sdk": patch
+---
+
+Fix add liquidity proportional query for cow-amm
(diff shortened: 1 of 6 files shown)


johngrantuk (approved):
Assume the fix has been discussed elsewhere so LGTM. 
Going forward, it would be useful to describe the changes/approach for stuff like this in the PR description to give a bit of context.

brunoguerios:
I agree. Will add the description for context anyway.

PR #403, opened 2024-08-28
Expect referenceAmount as AddLiquidityProportional input

### LINKED ISSUE #392: SDK - FE needs to query its own poolState to get bptOut out of referenceAmount provided as input
Current implementation of Add Liquidity Proportional expects `bptOut` as input (since it's an exactOut operation).
This causes a problem on the FE, because users input a referenceAmountIn in the UI.
Based on referenceAmountIn and a poolState (fetched on-chain by the FE), bptOut is estimated and then used as input to the SDK query.
SDK query will then rely on another poolState (fetched on-chain by the SDK) to perform that query.
Any eventual mismatch between both poolStates can cause the amountsIn calculated to be slightly different than the ones provided by the user.
This is an issue in the context of inputs being the max user balance, since any slightly higher amount will cause the transaction to revert.
A proposed solution to reduce the surface area for mistakes is to expect referenceAmountIn as the query input, so a single poolState is used for the whole calculation.

brunoguerios:
Closes #392 

This PR expands AddLiquidityProportional input to expect not only bptOut, but a referenceAmount from any pool token.

It will use on-chain state to calculate the bptOut corresponding to the referenceAmount provided and then perform the add liquidity query.

It's important to highlight that proportional math has inherent rounding issues, so we're rounding down in favor of leaving some dust behind instead of possibly returning an amount slightly higher, that could cause the transaction to revert.

Ps: for more context on the problem and the solution proposed, please refer to the attached issue.


MattPereira (approved):
Yo @brunoguerios amazing work on this one!!! Much better DX for SDK consumers :pray: 

My only suggestion would be that we consider renaming the examples like `addLiquidty.ts` could be `addLiquidityUnbalanced.ts` and `addLiquidityCowAmm.ts` could be `AddLiquidityProportional.ts`

src/entities/utils/getPoolStateWithBalancesV2.ts
```diff
@@ -0,0 +1,114 @@
...
+    http,
+} from 'viem';
+import { PoolType } from '@/types';
+import { HumanAmount } from '@/data';
+import { CHAINS, VAULT } from '@/utils';
+import { PoolState, PoolStateWithBalances } from '../types';
+import { getSortedTokens } from './getSortedTokens';
+import { vaultV2Abi } from '@/abi';
+
+type MulticallContract = {
```
johngrantuk, line 15: We've already using a type like this in `src/entities/utils/getPoolStateWithBalancesV3.ts` maybe worth reusing?
Also I see there's `getActualSupply` and `getTotalSupply` helpers in `src/utils/tokens.ts` that do similar things but I don't think are being used. Can you remove those?

test/lib/utils/removeLiquidityHelper.ts
```diff
@@ -509,17 +509,8 @@ export function assertTokenDeltas(
...
-            return Math.abs(parseInt(delta.toString()));
-        },
-    );
-    // Expected Delta for Recovery Exits has rounding errors because it relies
-    // on calculating proportional amounts off-chain instead of on-chain queries
-    balanceVsExpectedDeltas.forEach((diff) => {
-        expect(diff).to.be.lessThanOrEqual(5);
-    });
+
+    expect(txOutput.balanceDeltas).to.deep.eq(expectedDeltas);
```
johngrantuk, line 513: Nice to see the match!

test/lib/utils/addLiquidityHelper.ts
```diff
@@ -275,15 +274,13 @@ export function assertAddLiquiditySingleToken(
...
-        AddLiquidityQueryOutput,
-        'amountsIn' | 'bptIndex'
-    > = {
+    const expectedQueryOutput: Omit<AddLiquidityQueryOutput, 'bptIndex'> = {
         // Query should use same bpt out as user sets
         bptOut: TokenAmount.fromRawAmount(
             bptToken,
             addLiquidityInput.bptOut.rawAmount,
         ),
+        amountsIn: addLiquidityQueryOutput.amountsIn,
```
johngrantuk, line 283: If I understood correctly `expectedQueryOutput` is being tested against `addLiquidityQueryOutput` so maybe using values from `addLiquidityQueryOutput` itself isn't the best?

test/lib/utils/addLiquidityHelper.ts
```diff
@@ -340,21 +337,30 @@ export function assertAddLiquidityProportional(
...
-        'amountsIn' | 'bptIndex'
-    > = {
-        // Query should use same bpt out as user sets
-        bptOut: TokenAmount.fromRawAmount(
-            bptToken,
-            addLiquidityInput.bptOut.rawAmount,
-        ),
+    const expectedQueryOutput: Omit<AddLiquidityQueryOutput, 'bptIndex'> = {
+        // Query should use same referenceAmount as user sets
+        bptOut,
```
johngrantuk, line 362: Similar to above - seems like we're not really testing bptOut here?

test/lib/utils/addLiquidityHelper.ts
brunoguerios: This is just filling out content so I don't need create a separate scenario on `getCheck` without amountsIn.
There's no amountsIn provided as input that needs checking for Add Liquidity Single Token.
Do you think that's fine? Or is it too misleading?

test/lib/utils/addLiquidityHelper.ts
brunoguerios: We are comparing that to referenceAmount whenever reference amount is the bptOut. On other cases, I fill out with query amounts so it passes. Again, seems easier then adapting getCheck with/without bptOut depending on the input content.
If you think it's too confusing as is, I can spend some time on an alternative.

src/entities/utils/getPoolStateWithBalancesV2.ts
brunoguerios: `MulticallContract` is not really needed - I'll remove it.
`getActualSupply` and `getTotalSupply` are no longer being used because I've adapted in favor of multicalls - I'll remove them 👍

brunoguerios:
> Yo @brunoguerios amazing work on this one!!! Much better DX for SDK consumers 🙏
> 
> My only suggestion would be that we consider renaming the examples like
> 
> * `addLiquidty.ts` to `addLiquidityUnbalanced.ts`
> * `addLiquidityCowAmm.ts` to `addLiquidityProportional.ts`

I renamed them ✅ - but addLiquidityUnbalanced example is not working as expected. I created an [issue](https://github.com/balancer/b-sdk/issues/405) to keep track of it.

test/lib/utils/addLiquidityHelper.ts
johngrantuk: But as I understand it `expectedQueryOutput` is what we expect the query to return, even in the SingleToken case query, amountsIn contains an amount and we should really be testing that against an expect amount right?

johngrantuk (approved):
I still think the tests have issues but if we need to release to unblock FE it should be fine, can add a follow up issue.

test/lib/utils/addLiquidityHelper.ts
brunoguerios: AddLiquiditySingleToken is an exactOut transaction, so user input is actually bptOut and that is being tested against the query output.
But I agree that if that's causing confusion right now, it will definitely cause confusion on any eventual maintenance months from now.
I added a small refactor that I think will help better understand the nuances there.
Let me know what you think

brunoguerios:
> I still think the tests have issues but if we need to release to unblock FE it should be fine, can add a follow up issue.

Ok! I'll move on for now and we'll handle any future required changes in a new PR.

PR #404, opened 2024-08-28
Swap query support for hooks that require account

### LINKED ISSUE #398: Add optional account to swap query
Currently swap queries don't have an associated account. This means we can't query a swap that includes a hook that relies on specific account data (e.g. veBal balance discount fee).
Investigate following as possible solution:
* Add optional `account` param to `query`
* Where `await routerContract.simulate.query...` is called add `account` in params beside blockNumber
There might be a better way to do this.

MattPereira:
Closes #398

### Status: On Hold
- pending v3 contract change to add support account to queries so we dont have to psuedo query by simulating actual swap txs

### Summary
- Added optional `account` to `querySingleSwap` and `queryBatchSwap` ( for v3 only )
- Added example that uses custom swap path: `pnpm example ./examples/swaps/customSwap.ts`
    - Token and pool addresses come from [scaffold prod env](https://scaffold-balancerv3.vercel.app/pools?address=0x6D9656174205876897A9f526CCDcD3aE725ffEFF)
    - `account` is my dev wallet which holds the mock veBAL the hooks contract checks for


src/entities/swap/swaps/v3/index.ts
```diff
@@ -131,17 +147,30 @@ export class SwapV3 implements SwapBase {
...
-                    { blockNumber: block },
-                );
+            const { result } = account
+                ? await routerContract.simulate.swapSingleTokenExactOut(
+                      [
+                          this.swaps.pool,
+                          this.swaps.tokenIn,
+                          this.swaps.tokenOut,
+                          this.swaps.exactAmountOut,
+                          999999999999999999n, // maxAmountIn
```
MattPereira, line 157: Setting `maxAmoutnIn` like this is not ideal but not sure how to handle because `getLimitAmount` helper function requires  query results

src/entities/swap/swaps/v3/index.ts
```diff
@@ -97,30 +99,44 @@ export class SwapV3 implements SwapBase {
...
-                    { blockNumber: block },
-                );
+            const { result } = account
+                ? await routerContract.simulate.swapSingleTokenExactIn(
+                      [
+                          this.swaps.pool,
+                          this.swaps.tokenIn,
+                          this.swaps.tokenOut,
+                          this.swaps.exactAmountIn,
+                          0n, // minAmountOut
```
MattPereira, line 123: Same problem with `minAmountOut`

src/entities/swap/swaps/v3/index.ts
```diff
@@ -97,30 +99,44 @@ export class SwapV3 implements SwapBase {
 
         return this.isBatchSwap
             ? this.queryBatchSwap(client, block)
```
johngrantuk, line 101: Would likely need to do similar for batchSwap.

src/entities/swap/swaps/v3/index.ts
johngrantuk: You can use `MAX_UINT256` instead.
Not quite following what the issue is here (particularly in relation to getLimitAmount). Can you elaborate?

src/entities/swap/swaps/v3/index.ts
johngrantuk: This should be fine (see [Router](https://github.com/balancer/balancer-v3-monorepo/blob/e35fecb6634f7876f6c7b267bbf0b0e38bc008a6/pkg/vault/contracts/Router.sol#L1063) which does the same).

src/entities/swap/swaps/v3/index.ts
MattPereira: I noticed that `getLimitAmount()` is used to calculate the min/max amounts in/out, but it requires `queryOutput` which doesnt exist yet in the life cycle

But we are all good since you explained it's okay to just use `0n` and `MAX_UINT256`

src/entities/swap/swaps/v3/index.ts
MattPereira: Could use help understanding best methods for debugging errors with SDK. Like should I be using debugger with breakpoint and "Step into" ?

johngrantuk: I'd just go with whatever you normally use. I use a combo of console logs to narrow down the issue and debugger if i need to see a bunch of variables.

src/entities/swap/swaps/v3/index.ts
```diff
@@ -120,17 +123,31 @@ export class SwapV3 implements SwapBase {
...
-                        this.swaps.pool,
-                        this.swaps.tokenIn,
-                        this.swaps.tokenOut,
-                        this.swaps.exactAmountIn,
-                        DEFAULT_USERDATA,
-                    ],
-                    { blockNumber: block },
-                );
+            const { result } = account
+                ? // pseudo query fails if account has insufficient allowance or balance
```
johngrantuk, line 127: Don't have a big preference on ternary vs another method but did find it pretty tricky to notice the `?` with the comment inline like this.

MattPereira:
- Closing because #398 was closed to approach with new direction

PR #408, opened 2024-09-05
chore: Append vault and vault extension abis to any router interactions.

### LINKED ISSUE #370: Parse error signatures coming from SC
SC currently returns error signatures, such as `0xef029adf` that in fact could be parsed into `PoolNotInRecoveryMode`.

Here's Fabio's suggestion:
>I assume you're trying to interact with the pool via router and the vault throws that error? if you append the errors from the abi juani shared isn't viem able to show human-readable error messages?

Here's the json Juani shared:
[ivaulterrors.json](https://github.com/user-attachments/files/16363056/ivaulterrors.json)

Ps: not sure this approach is the best one (or that it even works), but I believe that getting human-readable error messages from SC would improve dev experience significantly. We could even consider making our input validator leaner and prioritize forwarding SC error messages more often within the SDK.

johngrantuk:
Closes #370 
Adding vault and vault extension abi provides the required error information to give more useful error messages. I tested manually and confirmed I got a readable message after. I don't see any issue in doing this but if anyone does please highlight!

### DIFF
--- a/.changeset/nine-pandas-judge.md
+++ b/.changeset/nine-pandas-judge.md
@@ -1,0 +1,5 @@
+---
+"@balancer/sdk": patch
+---
+
+Add vault and vault extension abis to any router interactions to enable decoded error messages.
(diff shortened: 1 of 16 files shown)


MattPereira (approved):
This is genius solution!!! :brain: 

Might also consider adding the permit2Abi as SDK consumers could mess that up as well ( speaking from expeirence :sweat_smile: )

brunoguerios:
What?! 🤯 
Which sorcery is this? 😂 
I'd love to better understand how this works.
Can you please to share more details or references to this?

Issue #411, opened 2024-09-12
Update to deploy 8


MattPereira:
To enable scaffold upgrades for the hookathon! :hook:

PR #412, opened 2024-09-12
Update to deploy 8

MattPereira:
WIP #411 

Just waiting for new mock pool addys


src/utils/constants.ts
```diff
@@ -205,6 +209,9 @@ export const PERMIT2: Record<number, Address> = {
 export const AUTHORIZER: Record<number, Address> = {
     [ChainId.SEPOLIA]: '0xa331d84ec860bf466b4cdccfb4ac09a1b43f3ae6',
 };
+
+export const ADMIN_OF_AUTHORIZER = '0x171C0fF5943CE5f133130436A29bF61E26516003'; // is this EoA chain agnostic?
```
MattPereira, line 213: From what I can tell, this EoA i capable of granting different roles, so maybe this is better? Are there multiple admins? Idk that much about OZ authorizer

https://sepolia.etherscan.io/address/0x171C0fF5943CE5f133130436A29bF61E26516003

test/v3/removeLiquidity.integration.test.ts
```diff
@@ -471,40 +476,73 @@ async function putPoolIntoRecoveryMode(
...
     authorizedAddress: Address,
 ) {
+    // get the actionId for the enableRecoveryMode function
+    const actionId = await client.readContract({
+        address: VAULT_ADMIN[chainId],
+        abi: vaultAdminAbi,
+        functionName: 'getActionId',
+        args: [toFunctionSelector('function enableRecoveryMode(address)')],
+    });
+
```
MattPereira, line 486: Think it is better to programmatically compute the actionId instead of storing it as a constant. Was a bit of a trip figuring out the actionId needed to be updated :sweat_smile:

src/utils/constants.ts
johngrantuk: I'm not sure of details here either. Maybe Mk will have an idea.

johngrantuk (approved):
LGTM

PR #417, opened 2024-09-23
Viem client type

agualis:
As a consumer, I want to have access to the type of the `client` parameter when calling methods like: 

`Relayer.signRelayerApproval` or `Permit2Helper.signAddLiquidityApproval` 

This PR refactors internal usages and exposes type:

```TS
export type ViemClient = Client & WalletActions & PublicActions;
```

### DIFF
--- a/src/utils/types.ts
+++ b/src/utils/types.ts
@@ -1,3 +1,3 @@
-import { Client, WalletActions, PublicActions } from 'viem';
+import { PublicActions, WalletClient } from 'viem';
 
-export type PublicWalletClient = Client & WalletActions & PublicActions;
+export type PublicWalletClient = WalletClient & PublicActions;
--- a/src/utils/types.ts
+++ b/src/utils/types.ts
@@ -1,0 +1,3 @@
+import { Client, WalletActions, PublicActions } from 'viem';
+
+export type ViemClient = Client & WalletActions & PublicActions;
--- a/src/utils/types.ts
+++ b/src/utils/types.ts
@@ -1,3 +1,3 @@
 import { Client, WalletActions, PublicActions } from 'viem';
 
-export type ViemClient = Client & WalletActions & PublicActions;
+export type PublicWalletClient = Client & WalletActions & PublicActions;
--- a/test/entities/swaps/v2/swapV2.integration.test.ts
+++ b/test/entities/swaps/v2/swapV2.integration.test.ts
@@ -8,10 +8,7 @@ import {
     parseEther,
     publicActions,
     walletActions,
-    Client,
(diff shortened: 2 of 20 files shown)


src/utils/types.ts
```diff
@@ -0,0 +1,3 @@
+import { Client, WalletActions, PublicActions } from 'viem';
+
+export type ViemClient = Client & WalletActions & PublicActions;
```
agualis, line 3: We could be more explicit with the type name using something like: 

`type PublicWalletViemClient` or so but, given that it is the only client that we are exposing, I would keep it shorter.

test/entities/swaps/v2/swapV2.integration.test.ts
```diff
@@ -44,7 +42,7 @@ const WETH = TOKENS[chainId].WETH;
 const vault = VAULT[chainId];
 
 describe('SwapV2', () => {
-    let client: Client & PublicActions & TestActions & WalletActions;
+    let client: ViemClient & TestActions;
```
MattPereira, line 45: I guess my lean would be to go with a type name of  `PublicWalletClient` for situations like this

As you mentioned, we are only using Viem so i dont think having it as part of type name adds much meaning

MattPereira (approved):
I'm mostly indifferent here with the explicit vs DRY trade-off, but I do support any changes that are helpful for our most important consumer (the v3 FE)

But plz wait for @johngrantuk to comment in case I'm missing some reason we shouldn't do this

src/utils/types.ts
agualis: Update: renamed to `PublicWalletClient`

agualis:
> I'm mostly indifferent here with the explicit vs DRY trade-off, but I do support any changes that are helpful for our most important consumer (the v3 FE)

For more context: the important thing from the consumer POV is exposing the type, if not, we need to do some [TS acrobatics to satisfy the compiler](https://github.com/balancer/frontend-v3/blob/fixUndefinedTokensInRemoveProvider/lib/modules/relayer/signRelayerApproval.ts#L9).

src/utils/types.ts
```diff
@@ -0,0 +1,3 @@
+import { Client, WalletActions, PublicActions } from 'viem';
+
+export type PublicWalletClient = Client & WalletActions & PublicActions;
```
johngrantuk, line 3: Could we replace this with:
```
export type PublicWalletClient = WalletClient & PublicActions;
```
I believe that is a bit tidier and should work.

johngrantuk (approved):
Apart from the one little change I suggested it all looks good to me. Thanks.

PR #423, opened 2024-10-01
fix: Add zero address sender to viem multicall calls to avoid issues with Nethermind/Geth nodes

agualis:
We found an issue with different RPC providers and the only solution is using `ZERO_ADDRESS` as the default account in certain multicall viem calls. 

More context: 
https://github.com/wevm/viem/issues/2792


johngrantuk (approved):
Interesting issue! Thanks for the update.

PR #429, opened 2024-10-08
Update to deploy 9

MattPereira:
You know what it is :rofl:


src/utils/constants.ts
```diff
@@ -199,19 +199,23 @@ export const COMPOSABLE_STABLE_POOL_FACTORY: Record<number, Address> = {
...
+    [ChainId.SEPOLIA]: '0x77eDc69766409C599F06Ef0B551a0990CBfe13A7',
 };
 
 export const BALANCER_BATCH_ROUTER: Record<number, Address> = {
-    [ChainId.SEPOLIA]: '0x0418001D0d68C71d0E391fE46dC7aFCe045f34A0',
+    [ChainId.SEPOLIA]: '0x16Cf31c5c4f92ad6185D583080C84FEeb6074c78',
+};
+
+export const BALANCER_COMPOSITE_LIQUIDITY_ROUTER: Record<number, Address> = {
+    [ChainId.SEPOLIA]: '0x89cA59Bc46c00D90C496Fc99f16668b00Dd6B5CC',
```
MattPereira, line 218: added this to help prepare for the upcoming boosting / nesting

src/abi/balancerCompositeLiquidityRouter.ts
MattPereira, line 1: this too

vitest.config.mts
```diff
@@ -24,7 +24,7 @@ export default defineConfig(({ mode }) => {
         },
         test: {
             testTimeout: 20_000,
-            hookTimeout: 30_000,
+            hookTimeout: 60_000,
```
MattPereira, line 27: As you can see by the commits on this PR, the create pool tests I just refactored will timeout on ~half the attempts. I think because I decided to move pool creation into the `beforeAll` hook. Let me know if this is bad practice. I just though it made sense to re-use same pool when testing deployment and initialization 

https://github.com/balancer/b-sdk/blob/976817520607576ce28ec6e93726510556222020/test/v3/createPool/stable/stable.integration.test.ts#L84-L88

johngrantuk: These tests have always been a pain! Not sure if we're missing something but haven't had time to check yet. Lets see how it goes for now.

johngrantuk (approved):
Thanks Matt.

PR #430, opened 2024-10-09
Test v3 swap e2e

mkflow27:
This pr adds e2e test for v3 swaps based on the already existing v3 integration tests with mock data. I additionally added some assertions which make sure the expected test case, matches what the api returns.

There are currently 2 tests in there which are being skipped. As well as boosted pool tests missing. Both of these can be reenabled/added once the sor is able to route through v3 stable pool liquidity.

I also had to fix the pnpm version due to not being able to run tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
         "vite-tsconfig-paths": "^4.2.3",
         "vitest": "^1.2.0"
     },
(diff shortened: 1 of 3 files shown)


johngrantuk:
Can reopen if neccessary.

PR #450, opened 2024-10-28
Feat boosted pool add and remove

mkflow27:
Created a seperate issue #455 for tracking one outstanding todo.


src/entities/addLiquidityBoosted/doAddLiquidityPropotionalQuery.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import { BALANCER_COMPOSITE_LIQUIDITY_ROUTER, CHAINS } from '@/utils';
+
+import { AddLiquidityProportionalInput } from '../addLiquidity/types';
+import { Address } from '@/types';
+
+import { balancerCompositeLiquidityRouterAbi } from '@/abi';
+
+export const doAddLiquidityProportionalQuery = async (
+    { rpcUrl, chainId, referenceAmount }: AddLiquidityProportionalInput,
```
johngrantuk, line 11: I would change this to accept user address and userData. (Not as optionals, index.ts can handle defaults)

src/entities/addLiquidityBoosted/doAddLiquidityUnbalancedQuery.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import { BALANCER_COMPOSITE_LIQUIDITY_ROUTER, CHAINS } from '@/utils';
+
+import { AddLiquidityUnbalancedInput } from '../addLiquidity/types';
+import { Address } from '@/types';
+
+import { balancerCompositeLiquidityRouterAbi } from '@/abi';
+
+export const doAddLiquidityUnbalancedQuery = async (
+    { rpcUrl, chainId, amountsIn }: AddLiquidityUnbalancedInput,
```
johngrantuk, line 11: Same as `doAddLiquidityProportionalQuery`

src/entities/addLiquidityBoosted/index.ts
```diff
@@ -0,0 +1,187 @@
...
+import { balancerCompositeLiquidityRouterAbi, balancerRouterAbi } from '@/abi';
+
+import { Hex } from '@/types';
+
+export class AddLiquidityBoostedV3 {
+    async query(
+        input: AddLiquidityInput,
+        poolState: PoolState | PoolStateWithUnderlyings,
+    ): Promise<AddLiquidityBaseQueryOutput> {
+        // Technically possible to pass singleToken adds here due to type
```
johngrantuk, line 37: Can change type to be 
`AddLiquidityUnbalancedInput
    | AddLiquidityProportionalInput;`

src/entities/addLiquidityBoosted/index.ts
```diff
@@ -0,0 +1,187 @@
...
+        // disallow it for this class
+        // TODO: Use input validator
```
johngrantuk, line 39: Couple of validations I can think of:
* Protocol version
* Input amounts are underlying

src/entities/addLiquidityBoosted/index.ts
```diff
@@ -0,0 +1,187 @@
...
+                    input.referenceAmount.rawAmount,
+                );
+                break;
+            }
+            case AddLiquidityKind.SingleToken: {
+                throw new Error('SingleToken not supported');
+            }
+        }
+
+        const output: AddLiquidityBaseQueryOutput = {
```
johngrantuk, line 95: In the nested PR I update to return userAddress and userData to buildCall can use that info.

src/entities/permit2Helper/index.ts
```diff
@@ -114,6 +114,50 @@ export class Permit2Helper {
...
+        ) {
+            throw new Error(
+                "Expirations length doesn't match amountsIn length",
+            );
+        }
+        const amounts = getAmountsCall(input);
+        const spender = BALANCER_COMPOSITE_LIQUIDITY_ROUTER[input.chainId];
+        const details: PermitDetails[] = [];
+
+        // unbalanced:
```
johngrantuk, line 140: Can remove all the debug stuff.

src/entities/removeLiquidityBoostedV3/doRemoveLiquidityProportionalQuery.ts
```diff
@@ -0,0 +1,33 @@
...
+import { BALANCER_COMPOSITE_LIQUIDITY_ROUTER, CHAINS } from '@/utils';
+
+import { Address } from '@/types';
+
+import { balancerCompositeLiquidityRouterAbi } from '@/abi';
+
+import { RemoveLiquidityProportionalInput } from 'src';
+
+export const doRemoveLiquidityProportionalQuery = async (
+    { rpcUrl, chainId, bptIn }: RemoveLiquidityProportionalInput,
```
johngrantuk, line 12: similar to adds, can update to accept address and userData

src/entities/addLiquidityBoosted/index.ts
```diff
@@ -0,0 +1,187 @@
...
+import { getValue } from '@/entities/utils/getValue';
+import { encodeFunctionData } from 'viem';
+import { balancerCompositeLiquidityRouterAbi, balancerRouterAbi } from '@/abi';
+
+import { Hex } from '@/types';
+
+export class AddLiquidityBoostedV3 {
+    async query(
+        input: AddLiquidityInput,
+        poolState: PoolState | PoolStateWithUnderlyings,
```
johngrantuk, line 35: Seems like there is something incorrect here? I don't think we can use `PoolState`?

src/entities/removeLiquidityBoostedV3/index.ts
```diff
@@ -0,0 +1,128 @@
...
+        const amountsOutInNumbers = await doRemoveLiquidityProportionalQuery(
+            input,
+            poolState.address,
+        );
+
+        // amountsOut are in underlying Tokens.
+        const amountsOut = amountsOutInNumbers.map((amount, i) => {
+            const token = new Token(
+                input.chainId,
+                poolState.tokens[i].underlyingToken.address,
```
johngrantuk, line 50: This is one of your build issues.

test/v3/addLiquidityNested/addLiquidityNestedV3.integration.test.ts
```diff
@@ -53,7 +53,7 @@ const usdcToken = new Token(chainId, USDC.address, USDC.decimals);
 const wethToken = new Token(chainId, WETH.address, WETH.decimals);
 const mainTokens = [wethToken, daiToken, usdcToken];
 
-describe.skip('V3 add liquidity nested test, with Permit2 direct approval', () => {
+describe('V3 add liquidity nested test, with Permit2 direct approval', () => {
```
johngrantuk, line 56: These skips are handled in my recent PR so you can just remove this diff to avoid conflicts.

test/v3/addLiquidityBoosted.integration.test.ts
johngrantuk, line 1: As discussed after your refactor ping me to let me know its good for review. I would recommend checking out the nested PR and maybe follow a similar approach which has two benefits imo: 1. Don't need to update all the existing test helpers which are convulted and hard to follow. 2. Makes the approval flow much easier to follow and confirm.

test/v3/removeLiquidityBoosted.integration.test.ts
johngrantuk, line 1: Same as add tests.

test/v3/addLiquidityBoosted.integration.test.ts
mkflow27: I have refactored the tests to not use the helpers and instead do the intermediary steps as part of the tests. I did not change anything in the way how I handle approvals as I think the way they are done in the tests here are fine.

test/v3/removeLiquidityBoosted.integration.test.ts
mkflow27: I utilise a helper here (which I think is ok) as a precondition for remove liq testing. Similarly I think the way I do approvals is ok. Compared to your tests, you do approvals right in the actualy tests, whereas I spread it.

src/entities/addLiquidityBoosted/doAddLiquidityPropotionalQuery.ts
mkflow27: I have addressed this as part of a larger refactor. It now allows to optionally supply `userData` & `userAddress` when querying the add/Remove classes and enforces values to be provided in the queries. Let me know what you think of that approach. It makes typing a bit overlapping as the same properties are either optionally or required.

src/entities/addLiquidityBoosted/doAddLiquidityUnbalancedQuery.ts
mkflow27: addresses in a larger refactor to use userData and userAddress

src/entities/addLiquidityBoosted/index.ts
mkflow27: The required state was actually dependend on the type. I now require to simply always pass a `PoolStateWithUnderlyings`. Only in the proportional adds was the type with underlying requires as the returned TokenAmounts need to have the pool's token underlyings but in the case of an unbalanced the returned TokenAmount is the BPT and all information was already shared in the input/poolstate.

src/entities/addLiquidityBoosted/index.ts
```diff
@@ -0,0 +1,187 @@
...
+                    return TokenAmount.fromRawAmount(
+                        new Token(input.chainId, t.address, t.decimals),
+                        t.rawAmount,
+                    );
+                });
+
+                break;
+            }
+            case AddLiquidityKind.Proportional: {
+                const exactAmountsInNumbers =
```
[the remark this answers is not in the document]
mkflow27: So the smart contracts will return the amounts of underlying tokens that the vault pulls in order to return an exactBpt out. So this variable just holds the numbers the SC returns. However the task is then below to return the TokenAmount as a combination of underlyingTokens & SC return values. I don't see how I would need to handle BPTs here. I need to handle the expected Tokens the Vault will pull in.

src/entities/permit2Helper/index.ts
mkflow27: addressed

src/entities/addLiquidityBoosted/index.ts
```diff
@@ -0,0 +1,187 @@
...
+        switch (input.addLiquidityKind) {
+            case AddLiquidityKind.Unbalanced: {
+                callData = encodeFunctionData({
+                    abi: balancerCompositeLiquidityRouterAbi,
+                    functionName: 'addLiquidityUnbalancedToERC4626Pool',
+                    args: [
+                        input.poolId,
+                        input.amountsIn.map((amount) => amount.amount),
+                        amounts.minimumBpt,
+                        !!input.wethIsEth,
```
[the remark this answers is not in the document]
mkflow27: addressed

src/entities/removeLiquidityBoostedV3/index.ts
```diff
@@ -0,0 +1,128 @@
...
+        const amounts = getAmountsCall(input);
+
+        const callData = encodeFunctionData({
+            abi: balancerCompositeLiquidityRouterAbi,
+            functionName: 'removeLiquidityProportionalFromERC4626Pool',
+            args: [
+                input.poolId,
+                input.bptIn.amount,
+                amounts.minAmountsOut,
+                !!input.wethIsEth,
```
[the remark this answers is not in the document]
mkflow27: addressed

src/entities/removeLiquidityBoostedV3/index.ts
mkflow27: build issues are addressed now

test/v3/addLiquidityNested/addLiquidityNestedV3.integration.test.ts
mkflow27: removed the dif.

src/entities/removeLiquidityBoostedV3/doRemoveLiquidityProportionalQuery.ts
mkflow27: Addressed

src/entities/addLiquidityBoosted/index.ts
mkflow27: I have addressed this for add- as well as remove liquidity

src/entities/addLiquidityBoosted/index.ts
```diff
@@ -0,0 +1,194 @@
...
+        // TODO: Use input validator
+        const bptToken = new Token(input.chainId, poolState.address, 18);
+
+        let bptOut: TokenAmount;
+        let amountsIn: TokenAmount[];
+        let tokenInIndex: number | undefined;
+
+        // Check if userAddress and userData were provided, and assign default values if not
+        if (!('userAddress' in input) || input.userAddress === undefined) {
+            // TODO: I think using a random address might be better than using the 0 address.
```
mkflow27, line 50: @johngrantuk I think defaulting to the 0 address might not be the right choice but rather a random address. What do you think?

src/data/providers/balancer-api/modules/pool-state/index.ts
```diff
@@ -39,6 +43,29 @@ export class Pools {
...
+          index
+          address
+          decimals
+          balance
+          underlyingToken {
+            address
+            decimals
+          }
+        }
+        dynamicData {
```
johngrantuk, line 63: I don't think we need this field?

src/entities/addLiquidityBoosted/index.ts
johngrantuk: I think zeroAddress is best but maybe you can explain why random is better?

src/entities/index.ts
```diff
@@ -24,3 +24,5 @@ export * from './types';
 export * from './utils';
 export * from './swap';
 export * from './swap/swaps/v2/auraBalSwaps';
+export * from './addLiquidityBoosted';
+export * from './removeLiquidityBoostedV3';
```
johngrantuk, line 28: Noticed one has V3 and one doesn't, lets use a consistent pattern, I'd say without V3?

src/data/providers/balancer-api/modules/pool-state/index.ts
mkflow27: You are right. I think I can actually remove the changes in the api, since I am not using the api in my tests, as we decided to use mock data. I'll remove it.

src/entities/addLiquidityBoosted/index.ts
mkflow27: My reasoning was that it is more likely for a query to return different data than what is expected as it is with a random address. This is probably an exotic scenario, so overall unlikely. So for readability / potential debugging purposes address 0 makes it easiert to work with

src/entities/addLiquidityBoosted/index.ts
```diff
@@ -0,0 +1,187 @@
...
+        // TODO: Use input validator
+        const bptToken = new Token(input.chainId, poolState.address, 18);
+
+        let bptOut: TokenAmount;
+        let amountsIn: TokenAmount[];
+        let tokenInIndex: number | undefined;
+        switch (input.kind) {
+            case AddLiquidityKind.Unbalanced: {
+                const bptAmountOut = await doAddLiquidityUnbalancedQuery(
+                    input,
```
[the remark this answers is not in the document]
mkflow27: I have added this and now fill the amounts in default values in case the user did not provide all the tokenAmounts he wants to join with.

src/entities/index.ts
mkflow27: Done. Renamed the folders to not have v3 anymore. For the classnames I found it easier to leave them with V3. Can also change. What do you think?

src/data/providers/balancer-api/modules/pool-state/index.ts
```diff
@@ -76,4 +103,24 @@ export class Pools {
...
+            variables: {
+                id: id.toLowerCase(),
+                chain: API_CHAIN_NAMES[this.balancerApiClient.chainId],
+            },
+        });
+        const poolStateWithUnderlyings: PoolStateWithUnderlyings = {
+            ...data.poolGetPool,
+            tokens: data.poolGetPool.poolTokens,
+            type: mapPoolType(data.poolGetPool.type),
+            //underlyings: data.poolGetPool.poolTokens.map((token) => {token}),
```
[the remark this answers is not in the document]
mkflow27: removed.

src/entities/addLiquidityBoosted/index.ts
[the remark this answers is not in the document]
johngrantuk: Separate issue to follow up on this: https://github.com/balancer/b-sdk/issues/455

PR #456, opened 2024-10-30
Feat boosted pool add and remove refactor

johngrantuk:


mkflow27 (approved):
Thanks John. Let's talk about the changes in our next 1:1

PR #458, opened 2024-10-31
Chore: remove unused totalShares field from PoolStateWithUnderlyings type

agualis:


johngrantuk (approved):
I originally left this in because I thought it was needed for `getBptAmountFromReferenceAmount` which we haven't implemented yet for boosted pools. But looking again it looks like the value is actually updated via an onchain call before calculation so should be good to remove. Thanks!

agualis:
> I originally left this in because I thought it was needed for `getBptAmountFromReferenceAmount` which we haven't implemented yet for boosted pools. But looking again it looks like the value is actually updated via an onchain call before calculation so should be good to remove. Thanks!

Good to read that we won't need it. The more we avoid dynamic state from the FE the better.

PR #474, opened 2024-11-04
Feat single step bpt swaps

mkflow27:
Adds support for First step being BPT Token when interacting with the batchRouter.

### DIFF
--- a/src/utils/constants.ts
+++ b/src/utils/constants.ts
@@ -211,7 +211,7 @@ export const BALANCER_ROUTER: Record<number, Address> = {
 };
 
 export const BALANCER_BATCH_ROUTER: Record<number, Address> = {
-    [ChainId.SEPOLIA]: '0x58b9e59a33AE8eF9b6B039118972a2C76F4c75D8',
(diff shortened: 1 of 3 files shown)


src/utils/constants.ts
```diff
@@ -211,7 +211,7 @@ export const BALANCER_ROUTER: Record<number, Address> = {
 };
 
 export const BALANCER_BATCH_ROUTER: Record<number, Address> = {
-    [ChainId.SEPOLIA]: '0x58b9e59a33AE8eF9b6B039118972a2C76F4c75D8',
+    [ChainId.SEPOLIA]: '0xB3AA271349c78FAe6180898dA1689C920573c937',
```
johngrantuk, line 214: Is this a new deployment? Didn't realise we're out of sync.

johngrantuk (changes requested):
Couple of small comments. Also looks like some tests are failing.

johngrantuk (approved):
Looks good, thanks! Can you add a changset before merging please (and mention there about the updated router address).

PR #486, opened 2024-11-08
Add proportional helper to calculate BPT from a reference amount for boosted pools

### LINKED ISSUE #455: Missing BPT amount calculation from reference amount in add liquidity boosted
Add the bpt amount calculation for add liquidity boosted, which is currently not supported.

brunoguerios:
Closes #455


test/v3/addLiquidity/addLiquidity.integration.test.ts
```diff
@@ -438,6 +438,7 @@ describe('add liquidity test', () => {
                 );
             });
         });
+        // FIXME: test proportional case on pool with >18 decimal tokens - it will fail
```
johngrantuk, line 441: Can we open an issue for this?

src/entities/utils/getBoostedPoolStateWithBalancesV3.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    const totalShares = outputs[0].result as bigint;
+    const balancesScale18 = outputs[1].result as bigint[];
+
+    const poolTokens = getSortedTokens(poolState.tokens, chainId);
+    const tokenBalances = poolTokens.map((token, i) =>
+        TokenAmount.fromScale18Amount(token, balancesScale18[i]),
+    );
+
+    const getUnderlyingBalancesContracts = poolState.tokens // TODO: not sorted
```
johngrantuk, line 60: Does this need to be sorted?

test/v3/addLiquidity/addLiquidity.integration.test.ts
brunoguerios: That not actually needed - I initially thought about testing the proportional helper by testing the add liquidity flow, but in the end, it made more sense to test the helper directly.
I'll remove the comment

src/entities/utils/getBoostedPoolStateWithBalancesV3.ts
brunoguerios: Daniel always said we should not blindly expect the API to return sorted tokens, so we should always sort locally by index. That's how it's handled everywhere else in the code (with `getSortedTokens` helper).
I added the comment so I wouldn't forget to do it before getting it ready for review - apparently I forgot 😅

brunoguerios: I fixed it and refactored the code to improve readability - let me know what you think

src/entities/utils/getBoostedPoolStateWithBalancesV3.ts
```diff
@@ -0,0 +1,139 @@
...
+                          ) as HumanAmount,
+                      },
+        })),
+        totalShares: formatEther(totalShares) as HumanAmount,
+    };
+    return poolStateWithUnderlyingBalances;
+};
+
+const getUnderlyingBalances = async (
+    sortedTokens: import('/Users/bruninho/Development/Balancer/b-sdk/src/entities/types').PoolTokenWithUnderlying[],
```
johngrantuk, line 72: Something weird going on here?

src/entities/utils/proportionalAmountsHelpers.ts
```diff
@@ -127,3 +129,41 @@ export const getBptAmountFromReferenceAmount = async (
...
+ * @param input
+ * @param poolState
+ * @returns
+ */
+export const getBptAmountFromReferenceAmountBoosted = async (
+    input: AddLiquidityBoostedProportionalInput,
+    poolStateWithUnderlyings: PoolStateWithUnderlyings,
+): Promise<InputAmount> => {
+    let bptAmount: InputAmount;
+    if (input.referenceAmount.address === poolStateWithUnderlyings.address) {
```
johngrantuk, line 145: Should use `isSameAddress` here?

src/entities/utils/getBoostedPoolStateWithBalancesV3.ts
johngrantuk, line 1: I think a few comments describing whats going would help a reader parse the logic a bit quicker.

johngrantuk (changes requested):
Just a few comments but its looking good.

src/entities/utils/getBoostedPoolStateWithBalancesV3.ts
brunoguerios: done ✅

src/entities/utils/getBoostedPoolStateWithBalancesV3.ts
```diff
@@ -0,0 +1,139 @@
...
+        );
+    }
+
+    const underlyingBalances = underlyingBalanceOutputs.map(
+        (output) => output.result as bigint,
+    );
+    return underlyingBalances;
+};
+
+const getTokenBalancesAndTotalShares = async (
```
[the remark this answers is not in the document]
brunoguerios: done ✅

src/entities/utils/proportionalAmountsHelpers.ts
brunoguerios: done ✅

src/entities/utils/getBoostedPoolStateWithBalancesV3.ts
brunoguerios: done ✅

test/v3/addLiquidityBoosted/addLiquidityPartialBoosted.integration.test.ts
```diff
@@ -80,91 +80,195 @@ describe('V3 add liquidity partial boosted', () => {
...
+
+        test('query with underlying', async () => {
+            const queryOutput = await addLiquidityBoosted.query(
+                addLiquidityInput,
+                partialBoostedPool_USDT_stataDAI,
             );
+            expect(queryOutput.protocolVersion).toEqual(3);
+            expect(queryOutput.bptOut.token).to.deep.eq(parentBptToken);
+            expect(queryOutput.bptOut.amount > 0n).to.be.true;
+            expect(queryOutput.amountsIn[0]).to.deep.eq(referenceTokenAmount);
```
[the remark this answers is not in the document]
brunoguerios: done ✅

test/entities/swaps/v2/auraBalSwaps/auraBal.integration.test.ts
brunoguerios, line 1: This is an attempt to fix an unrelated test that was intermittently failing.

MattPereira: Nice creating the test client just once in the before all hook is good improvment :trophy:

test/v3/addLiquidityBoosted/addLiquidityPartialBoosted.integration.test.ts
```diff
@@ -52,16 +54,20 @@ describe('V3 add liquidity partial boosted', () => {
...
         TokenAmount.fromHumanAmount(daiToken, '2'),
     ];
 
     beforeAll(async () => {
         // setup chain and test client
-        ({ rpcUrl } = await startFork(ANVIL_NETWORKS.SEPOLIA));
+        ({ rpcUrl } = await startFork(
+            ANVIL_NETWORKS.SEPOLIA,
+            undefined,
+            7010800n,
```
MattPereira, line 69: Is the block number fixed here to protect against external changes to pool liquidity that could cause test to fail?

Looks like the `addLiquidityBoosted.integration.test.ts` file does not specify a block number when starting the fork. Is there some reason that is okay there but not here? 

Or is it better practice to always fix the blockNumber the fork starts? Or is that not worth the trade off of having to update block numbers each time a new version of v3 is deployed? 

Maybe could make sense to centralize the block all v3 tests run at using a constants file?

src/entities/utils/getPoolStateWithBalancesV3.ts
MattPereira, line 1: Soooo clean and ez to read :relaxed:

MattPereira (approved):
Matou a pau! :wood: ✨

test/entities/swaps/v2/auraBalSwaps/auraBal.integration.test.ts
johngrantuk: Thanks for trying to address this! Whats your thoughts on the reason it was intermittently failing?

johngrantuk (approved):
Looks good! Once we're happy with the final decision on test block number lets release and let the FE guys know so it can be used as part of the QA testing that is due to start soon.

test/entities/swaps/v2/auraBalSwaps/auraBal.integration.test.ts
brunoguerios: My assumption was that the failure came from parallel threads not being handled properly. Since tests were reusing the same fork, they could be affecting each other's results.
The idea of forcing a snapshot and keeping each test on its own fork sounded like something that would prevent this from happening.
But then again, I'm not completely sure this is the case and that's why I mentioned it was an "attempt" 😅 
We can dig deeper in case the issue starts happening again 🤷‍♂️

test/v3/addLiquidityBoosted/addLiquidityPartialBoosted.integration.test.ts
brunoguerios: Yeah, I had to fix the blockNumber due to how tests were setup (see line 220).
On this addLiquidityPartialBoosted we're testing the queryOutput separated from the actual transaction, so we can't compare to token balance delta as we do on other files.
For add liquidity unbalanced that's ok, because we can compare queryOutputs with amountsIn provided as input, but for add liquidity proportional, we don't have that.
So, if we want to check queryOutput against all amountsIn, we have to provide a known/hardcoded value, which requires blockNumber to be fixed.
With that being said, I usually prefer to avoid fixing blockNumbers as well.
We could refactor addLiquidityPartialBoosted to follow the same pattern as addLiquidityBoosted and this would be solved.
What do you think?

PR #489, opened 2024-11-11
Refactor nestedPoolState logic to fetch mainTokens from api

### LINKED ISSUE #488: Refactor nested pool data schema to follow api update
Api is deprecating `allTokens`, `displayTokens` and `tokens` in favor of `pooltokens`.

Nested Pool State needs to replace `allTokens` + `isMainToken` logic.

@franzns sugestion:
> You can check wether hasNestedPool is null, then it is a mainToken. If you look for the phantom BPT, you compare token and pool address

brunoguerios:
Closes #488

### DIFF
--- a/src/data/providers/balancer-api/modules/nested-pool-state/index.ts
+++ b/src/data/providers/balancer-api/modules/nested-pool-state/index.ts
@@ -249,24 +249,9 @@ export function mapPoolToNestedPoolStateV2(pool: PoolGetPool): NestedPoolState {
     });
 
     // mainTokens are pool tokens filtering out nested pools and phantomBPTs
-    const mainTokens = pool.poolTokens.flatMap((t) => {
-        if (t.nestedPool) {
-            const nestedPool = t.nestedPool;
-            return nestedPool.tokens
-                .filter((t) => t.address !== nestedPool.address) // remove phantomBPT
-                .map((t) => {
-                    return {
-                        address: t.address,
-                        decimals: t.decimals,
-                    };
-                });
-        }
-
-        return {
-            address: t.address,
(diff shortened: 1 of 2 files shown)


src/data/providers/balancer-api/modules/nested-pool-state/index.ts
```diff
@@ -258,14 +248,25 @@ export function mapPoolToNestedPoolStateV2(pool: PoolGetPool): NestedPoolState {
...
-        .filter((t) => t.isMainToken)
-        .map((t) => {
-            return {
-                address: t.address,
-                decimals: t.decimals,
-            };
-        });
+    // mainTokens are pool tokens filtering out nested pools and phantomBPTs
+    const mainTokens = pool.poolTokens.flatMap((t) => {
+        if (t.nestedPool) {
```
franzns, line 253: Think you need to filter the phantomBPT first, otherwise you get the mainTokens of the parent pool twice. 

Also think about duplicated tokens. A usdc/bb-a-usd pool would have 2x usdc, do you need it twice or just once?

brunoguerios: You're right about filtering out phantomBPT from the parent pool - I'll fix that.
But we don't have to worry about duplicate tokens as that's not supported by the SDK - it will fail in the validator. There shouldn't exist such a pool composition.

brunoguerios: I changed the logic a bit to make it simpler and easier to read/understand

franzns: Looks good!

PR #490, opened 2024-11-11
Fix getPoolStateWithBalancesV3 with less 18 decimals tokens

### LINKED ISSUE #487: getPoolStateWithBalancesV3 won't work for tokens with <18 decimals
This happens because `getCurrentLiveBalances` returns `scale18` amounts, while the SDK is handling as if they were `raw` amounts

I'm assuming there isn't a similar SC method that returns raw amounts instead, so we should handle scaling in the SDK, but it would be better to double check.

brunoguerios:
Closes #487

### DIFF
--- a/.changeset/pink-phones-suffer.md
+++ b/.changeset/pink-phones-suffer.md
@@ -1,0 +1,5 @@
+---
+"@balancer/sdk": patch
+---
+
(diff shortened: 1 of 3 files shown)


johngrantuk (approved):
Good catch!

Issue #339, opened 2024-06-28
Update docs SDK examples/guides


johngrantuk:
Once we have more stable codebase (mainly around permit/permit2) we need to update docs examples/guides to use most recent SDK interfaces.

MattPereira:
Some realizations:

- Permit2 examples will require a `process.env.PRIVATE_KEY` in order to `.signTypedData`
- Should we have seperate `swapV2` and `swapV3` examples since only v3 uses permit2?
- Permit2 examples will have to be on Sepolia cus that's the only chain Router is deployed
- The token pair from scaffold prod does not get picked up by SOR paths fetched from API

johngrantuk:
Thinking about this a bit more how about these separate examples as a fresh approach?

- How to fetch a swap
  - Fetches from api and shows how to re-query
  - No build call
- Swap With Approvals
  - Run on a fork (similar to removeLiquidity example)
  - Info at top explaining this is only way V2 works, mention signature alternative for V3
  - Not using API - use a hard coded V2 path
  - Approval process goes through function exported from separate example (see below)
  - use buildCall
- Swap With Signature
  - Run on a fork (similar to removeLiquidity example)
  - Info at top explaining this is V3 only and why Permit2/Signatures can be used
  - Not using API - use a hard coded path
  - Approval process goes through function exported from separate example (see below)
  - use buildCallWithPermit2
- Approvals
  - Shows the various methods of approving for V2/V3:
    - V2 approves vault
    - V3 with normal approvals, approve permit2, approve router
    - V3 with signature, approve permit2, signature
  - Exports each method as a function that can be used in other examples

MattPereira:
Looks like the approvals are currently being abstracted away in `makeForkTx > forkSetup > approveTokens`. 

We want to bring those to the surface and be more explicit right?

MattPereira:
Definitely agree with modularizing the examples. Might be good to outline the folder/file structure as starting point.

Not sure best way to handle custom vs smart path given the different approval options. Kinda gross to repeat

Feel free to edit:

- 📁 examples
  - 📁 swaps
    - 📄 query.ts
    - 📁 withApproval
      - 📄 customPath.ts
      - 📄 smartPath.ts
    - 📁 withSignature
      - 📄 customPath.ts
      - 📄 smartPath.ts
  - 📁 approvals
    - 📁 v2
    - 📁 v3

MattPereira:
[...]

PR #498, opened 2024-11-22
Add Permit and Permit2 signature docs

### LINKED ISSUE #369: Permit Signatures - Document relevant info
Relevant info to be shared regarding using Permit and Permit2 signatures in the SDK:
- Balancer v3 allows token approval flow to be handled through Permit and Permit2.
- It can be done through direct approvals (ERC20 or Permit2) or through signatures.
- To facilitate using signatures, SDK:
   - includes buildCallWithPermit that expects Permit signatures for Remove Liquidity operations
   - includes buildCallWithPermit2 that expects Permit2 signatures for Add Liquidity and Swap operations
   - provides helpers to sign Permit and Permit2 signatures
- all permit signatures operations (swaps, add/remove liquidity) support native asset as input/output (i..e ETH)

brunoguerios:
Closes #369


README.md
```diff
@@ -700,4 +728,94 @@ calculateProportionalAmounts(
...
+| nonces (optional) | number[] | Nonces for each token |
+| expirations (optional) | number[] | Expirations for each token |
+
+**Returns**
+
+```typescript
+Promise<Permit2>;
+```
+
+[Permit2](./src/entities/permit2Helper/index.ts#L35) - Permit2 object with metadata and encoded signature
```
MattPereira, line 774: I like this! Good balance between brevity and detail

If the plan is to detail each of the `Permit2Helper` methods, this is a great blueprint :muscle:

examples/addLiquidity/addLiquidityWithPermit2Signature.ts
```diff
@@ -0,0 +1,172 @@
...
+                address: a.address,
+                slot: getSlot(chainId, a.address),
+                rawBalance: a.rawAmount,
+            })),
+            client,
+        },
+        [...amountsIn.map((a) => a.address), pool.address],
+        call.protocolVersion,
+        approveOnPermit2,
+    );
```
MattPereira, line 96: Nice move adding the flag that prevents the default cascade of approvals by `makeForkTx` so that we can know the example is working properly!

I do have a larger suggestion for team to consider that I will bring up during standup:

`makeForkTx` may confuse less experienced devs who might be distracted by token slots or not fully grasp how the call object the SDK creates is used to send the tx. 

This may be a stretch that requires a fair bit of refactor, but perhaps rolling a simplified fork setup for examples could be an option where we use the default anvil account #0 to
1. wrap ETH into wETH
2. swap wETH into BAL
3. add liquidity to wETH/BAL pool
4. any necessary setup for boosted or nested joins / removes

From there, we should be able to focus examples on just the code needed for each operation and it would be easier to explain a simplified fork setup in some detail

Not suggesting this as part of this PR and not too attached to this idea, just a suggestion that could enable examples to be a little more straightforward

P.S. John had [this idea for examples](https://github.com/balancer/b-sdk/issues/339#issuecomment-2337592756) a while back. Not sure if it applies to add / remove, but wanted to raise awareness

brunoguerios: I'll keep as is for now, since we're evaluating different approach for examples on a separate PR.

README.md
MattPereira: i'm on the fence as well. might not be a bad idea to hope for the best and only add more explicit docs for each method if we get complaints

examples/addLiquidity/addLiquidityUnbalanced.ts
MattPereira, line 1: Attempt to run this script seems to result in the following error:

```
[ b-sdk ] $ pnpm example ./examples/addLiquidity/addLiquidityUnbalanced.ts

> @balancer/sdk@0.32.2 example /home/matthu/Desktop/Balancer/b-sdk
> npx tsx ./examples/lib/executeExample.ts "./examples/addLiquidity/addLiquidityUnbalanced.ts"

checking rpcUrl 8945 {}
🛠️  Starting anvil { port: 8945, forkBlockNumber: 7010800n }
/home/matthu/Desktop/Balancer/b-sdk/src/data/providers/balancer-api/modules/pool-state/index.ts:54
            ...data.poolGetPool,
                    ^


TypeError: Cannot read properties of null (reading 'poolGetPool')
    at Pools.fetchPoolState (/home/matthu/Desktop/Balancer/b-sdk/src/data/providers/balancer-api/modules/pool-state/index.ts:54:21)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async addLiquidityExample (/home/matthu/Desktop/Balancer/b-sdk/examples/addLiquidity/addLiquidityUnbalanced.ts:82:23)
    at async runAgainstFork (/home/matthu/Desktop/Balancer/b-sdk/examples/addLiquidity/addLiquidityUnbalanced.ts:47:18)
    at async executeExample (/home/matthu/Desktop/Balancer/b-sdk/examples/lib/executeExample.ts:6:5)

Node.js v20.17.0
 ELIFECYCLE  Command failed with exit code 1
 ```

MattPereira: Tried moving the forkBlockNumber but still fails

MattPereira: Recreated the query and looks like the response from API is coming up dry

image
image

MattPereira: Maybe the API is on deploy11 but the pool is from deploy10? 

Heres a [tenderly simulation](https://www.tdly.co/shared/simulation/8247adf2-f1be-40d0-aca4-45c0457ec30d) at deploy11 vault explorer that shows pool is not registered

MattPereira: Yeah looks like this PR is on deploy 10 and API is on deploy 11

Examples run fine after merging main into this PR and updating with a deploy 11 pool address

brunoguerios: Thanks for checking that!
You're right - sorry - I had this open before updating to deploy11. 
I'll update the PR over here.

brunoguerios: done ✅

PR #502, opened 2024-11-27
Update to Deploy 11

### LINKED ISSUE #499: Update to Deploy11

brunoguerios:
Closes #499

### DIFF
--- a/examples/createAndInitPool/createAndInitPoolV3.ts
+++ b/examples/createAndInitPool/createAndInitPoolV3.ts
@@ -1,0 +1,314 @@
+/**
+ * Quick script to deploy partially boosted pool we need for testing
+ *
+ * Run with:
+ * pnpm example ./examples/createAndInitPool/createAndInitPoolV3.ts
+ */
+
+import { config } from 'dotenv';
+config();
+
+import {
+    // createTestClient,
+    http,
+    publicActions,
+    // walletActions,
+    zeroAddress,
+    parseUnits,
+    createWalletClient,
+    getContract,
+    PrivateKeyAccount,
+} from 'viem';
+import { privateKeyToAccount } from 'viem/accounts';
+import {
+    stablePoolFactoryAbi_V3,
+    CreatePoolV3StableInput,
+    CreatePool,
+    PoolType,
+    ChainId,
+    CHAINS,
+    InitPoolDataProvider,
+    InitPool,
+    InitPoolInput,
+    TokenType,
+    PERMIT2,
+    BALANCER_ROUTER,
+    permit2Abi,
+    erc20Abi,
+    MaxAllowanceExpiration,
+    PermitDetails,
+    Permit2Batch,
+    MaxSigDeadline,
+    AllowanceTransfer,
+    balancerRouterAbi,
+    PublicWalletClient,
+} from 'src';
+// import { startFork, ANVIL_NETWORKS } from 'test/anvil/anvil-global-setup';
+import { findEventInReceiptLogs } from 'test/lib/utils/findEventInReceiptLogs';
+// import { makeForkTx } from 'examples/lib/makeForkTx';
+// import { getSlot } from 'examples/lib/getSlot';
+
+const SWAP_FEE_PERCENTAGE_DECIMALS = 16;
+const AAVE_FAUCET_USDT = {
+    address: '0xaa8e23fb1079ea71e0a56f48a2aa51851d8433d0' as `0x${string}`,
+    decimals: 6,
+};
+const STATA_ETH_DAI = {
+    address: '0xDE46e43F46ff74A23a65EBb0580cbe3dFE684a17' as `0x${string}`,
+    rateProvider: '0x22db61f3a8d81d3d427a157fdae8c7eb5b5fd373' as `0x${string}`,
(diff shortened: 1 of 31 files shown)


test/entities/swaps/v3/swapV3.integration.test.ts
```diff
@@ -221,7 +215,9 @@ describe('SwapV3', () => {
         snapshot = await client.snapshot();
     });
 
-    describe('query method should return correct updated', () => {
+    // TODO: double check if comparing query outputs against balanceDeltas isn't redundant with query tests
+    // if yes, we should be able to remove query tests (and avoid relying on a fixed blockNumber)
+    describe.skip('query method should return correct updated', () => {
```
brunoguerios, line 220: Highlighting this as it might be the one that sparks a bit of discussion.
Main question is: 
> do we need to test queryOutputs against hardcoded amounts?

I have a feeling that testing `queryOutput` against `balanceDeltas` from the actual transaction should be enough.
If we manage to remove these, we won't need to rely on a fixed blockNumber for these tests.

test/v3/addLiquidityBoosted/addLiquidityPartialBoosted.integration.test.ts
```diff
@@ -50,7 +50,8 @@ const parentBptToken = new Token(
 const usdtToken = new Token(chainId, USDT.address, USDT.decimals);
 const daiToken = new Token(chainId, DAI.address, DAI.decimals);
 
-describe('V3 add liquidity partial boosted', () => {
+// TODO: pending test pool to be created/initialized
+describe.skip('V3 add liquidity partial boosted', () => {
```
brunoguerios, line 54: I already pinged @mendesfabio - as soon as the test pool is created/initialized, we should be able to update addresses and hopefully no other changes will be required

test/v3/priceImpact/priceImpact.V3.integration.test.ts
```diff
@@ -176,7 +177,8 @@ describe('PriceImpact V3', () => {
         });
     });
 
-    describe('Nested pool', () => {
+    // FIXME: zeroOutDeltas is swapping a huge amount (~200 WETH) and hitting MaxInRatio
+    describe.skip('Nested pool', () => {
```
brunoguerios, line 181: This one is indeed not working, but I'd suggest we fix this in a separate PR, so we unblock the deploy update. Let me know what you think and I'll create a ticker over here

test/v3/addLiquidityBoosted/addLiquidityPartialBoosted.integration.test.ts
MattPereira: Test is now passing after updating with identical partially boosted pool that I deployed with [this script](https://github.com/balancer/b-sdk/blob/deploy-11/examples/createAndInitPool/createAndInitPoolV3.ts)

test/v3/priceImpact/priceImpact.V3.integration.test.ts
MattPereira: I agree with kicking this down the road

examples/createAndInitPool/createAndInitPoolV3.ts
MattPereira, line 1: This is still kinda messy and would benefit from Permit2helper init method, but seems like decent pinch option for deploying v3 pools and could start discussion about examples that ready to send transactions to live network

No worries if you prefer to dump it for this PR, just thought I'd offer the option to decide

test/anvil/anvil-global-setup.ts
```diff
@@ -66,7 +66,7 @@ export const ANVIL_NETWORKS: Record<NetworksWithFork, NetworkSetup> = {
         rpcEnv: 'SEPOLIA_RPC_URL',
         fallBackRpc: 'https://sepolia.gateway.tenderly.co',
         port: ANVIL_PORTS.SEPOLIA,
-        forkBlockNumber: 7164489n,
+        forkBlockNumber: 7167280n, // Matt moved this blockNo after deploying new partial boosted pool
```
MattPereira, line 69: Had to move this block number to get the `addLiquidityPartialBoosted.integration.test.ts` test to pass, but as you can see it causes `priceImpact.V3.integration.test.ts` to fail. Maybe there was a trade in the pool?

 I'm not sure if it is okay to move the hardcoded values to fix
 
image

test/entities/swaps/v3/swapV3.integration.test.ts
johngrantuk: Keeping in mind that after we launch there will be no new deployments and fixed block number will be permanent so it's probably not going to be that important soon. The main benefit I see is it potentially catches if changes break anything unexpectedly (e.g. results should always be the same). Its maybe not that important, especially if we're definitely checking that the query result is at least > 0 in `assertSwapExactIn/Out` but I'm not 100% sure that is the case and it would need to be confirmed first.

examples/createAndInitPool/createAndInitPoolV3.ts
johngrantuk: Can leave it in for now but should be tidied as part of wider example changes.

PR #509, opened 2024-11-29
Fix pool init error for pool creation UI

MattPereira:
I can't seem to figure out why SDK query for poolState is failing. See[ code comment below](https://github.com/balancer/b-sdk/pull/509#discussion_r1863946915)

I have confirmed using the vaultExplorer contract that the pool is registered and `getPoolTokens` works there. 

At this point, I'm unsure if bug is in SDK or pool creation UI

### Resources
- pool address: `0xF5A25990da505a9d4C3E0d876f3951E9EDF9ABC3`
- vault address: `0xBC582d2628FcD404254a1e12CB714967Ce428915`
- vault explorer address: `0xa9F171e84A95c103aD4aFAC3Ec83810f9cA193a8`
- [Tenderly call](https://www.tdly.co/shared/simulation/10b62509-7b2d-4e39-bbd9-92d9378cd0fc)  at vault with vaultExtension ABI

### DIFF
--- a/src/utils/tokens.ts
+++ b/src/utils/tokens.ts
@@ -1,6 +1,6 @@
 import { Address, Client, Hex, PublicClient, getContract } from 'viem';
-import { vaultV2Abi, vaultExtensionAbi_V3 } from '../abi';
-import { VAULT, VAULT_V3 } from './constants';
+import { vaultV2Abi, vaultExplorerAbi_V3 } from '../abi';
+import { VAULT, VAULT_EXPLORER_V3 } from './constants';
 
 export async function getTokenDecimals(
     tokenAddress: Address,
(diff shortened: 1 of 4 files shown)


src/utils/tokens.ts
```diff
@@ -68,16 +68,18 @@ export async function getPoolTokensV3(
...
         });
-        return (await vaultV3.read.getPoolTokens([poolAddress])) as Address[];
+        return (await vaultExplorer.read.getPoolTokens([
+            poolAddress,
+        ])) as Address[];
     } catch (e) {
         console.warn(e);
         throw new Error(
-            `Error: Unable to get pool tokens using this pool address: ${poolAddress}`,
+            `Unable to get pool tokens using this pool address: ${poolAddress}`,
```
MattPereira, line 82: I don't understand why this stopped working, but I'm guessing vault explorer will work?

Will vault explorer be deployed for live networks?

brunoguerios: I also don't understand why vault call would fail here and vaultExplorer would succeed 🤔

MattPereira:
Closing because pretty sure the problem was related to bad rpc url. Thanks @brunoguerios ! :pray:

PR #510, opened 2024-12-01
Refactor swap examples

### LINKED ISSUE #339: Update docs SDK examples/guides

MattPereira:
Closes #339

My suggestion for re-shaping examples. Feel free to critique, modify, ect.

### Summary
- setup simple example fork hoping to make process easier to follow along and modify to prep for other examples
- split swap queries into discrete / re-usable examples
- `pnpm example ./examples/swaps/swapV2.ts` uses SOR path with old school v2 approval of vault
- `pnpm example ./examples/swaps/swapV3.ts` uses custom path with permit2 approval

### Console Logs
image


examples/swaps/swapV3.ts
```diff
@@ -0,0 +1,118 @@
...
+    // Choose chain id to start fork
+    const chainId = ChainId.SEPOLIA;
+    const { client, rpcUrl, userAccount } = await setupExampleFork({ chainId });
+
+    // TODO: Fix query revert, see tenderly simulation: https://www.tdly.co/shared/simulation/f0b0a1de-4e88-4b8d-bf90-c25c4d2145ec
+    // Query swap results before sending transaction
+    const { swap, queryOutput } = await queryCustomPath({
+        rpcUrl,
+        chainId,
+        pools: ['0xA8c18CE5E987d7D82ccAccB93223e9bb5Df4A3c0'] as Address[], // https://test.balancer.fi/pools/sepolia/v3/0xa8c18ce5e987d7d82ccaccb93223e9bb5df4a3c0
```
MattPereira, line 33: Still working on debugging this :sweat_smile: 

The pools listed on v3 FE are deploy 11 right?
https://test.balancer.fi/pools?networks=SEPOLIA&skip=0&textSearch=bal&protocolVersion=3

brunoguerios (commented):
I like this new flow!

Splitting query/transaction examples + reusing them is nice!
And the fact that you're not relying on manipulating storage slots for adding token balance to test accounts feels better as well.
setupExampleFork will get a bit more complex once you swap + add liquidity to prepare for other examples, but I'd say it's quite straightforward to follow.

One thing that came to my mind is that I'm used to leverage examples as debugging tools, where I simply adapt chains/pools/tokens to a different scenario to see if everything is working as expected. This new flow is a bit less flexible in that sense, but I'd say it makes sense to create separate debug files and use those instead of trying to reuse examples for that purpose.

brunoguerios (approved):
I think this works great! 👏 
I'd say we should follow this pattern with other actions (add/remove/etc..)
And it would be nice to sync so we make sure we're covering everything we need to provide good DX.

PR #514, opened 2024-12-03
Add native token support to AddLiquidityNested and AddLiquidityBoosted

### LINKED ISSUE #439: V3 Full Native Support
After deploy11 we can investigate what is missing and implement:
* Nested
* Boosted

brunoguerios:
Closes #439 

Main changes:
- Add support for native token AddLiquidityNested
- Add support for native token AddLiquidityBoosted
- Add support for native token RemoveLiquidityNested
- Add support for native token RemoveLiquidityBoosted

Notes:
- Removed native token variations from boostedPool integration tests because we don't have a testPool with "wrapped" WETH
    - I believe we should be covered with partialBoostedPool integration tests only, but let me know what you think.
- Replaced partialBoosted test pool on the integration test with one that has WETH so we can test `wethIsEth = true`
- Refactored addLiquidityBoostedHelper so we can reuse between integration tests
- Refactored removeLiquidityNestedHelper so we can reuse between integration tests
- Fixed priceImpact integration test blockNumber so we don't need to keep updating hardcoded amounts


test/v3/addLiquidityBoosted/addLiquidityBoosted.integration.test.ts
```diff
@@ -172,9 +172,6 @@ describe('Boosted AddLiquidity', () => {
                         addLiquidityBuildCallOutput.minBptOut.amount,
                 ).to.be.true;
             });
```
johngrantuk, line 174: Just to confirm - we're happy native tests are covered enough in addLiquidityPartialBoosted.integration.test.ts?

johngrantuk (approved):
Looks good!

test/v3/addLiquidityBoosted/addLiquidityBoosted.integration.test.ts
brunoguerios: I'd say that from the SDK perspective we should be covered because they reuse the same implementation 👍 
In any case, I'd say we can ship as is and ask SC team to create a test pool with wrapped WETH so we can test on the UI.

test/lib/utils/addLiquidityBoostedHelper.ts
```diff
@@ -130,3 +136,170 @@ export const assertTokenMatch = (
         expect(a.decimals).to.eq(tokenReturned[i].decimals);
     });
 };
+
+export type AddLiquidityBoostedTxInput = {
+    client: PublicWalletClient & TestActions;
```
MattPereira, line 141: Could including `TestActions` type here cause problem for SDK consumer's who want to send tx's to live networks?

brunoguerios: This type is used only within tests, so it should be ok.

MattPereira: Err nvm i forgot SDK abstracts away complexity of boosted and only asks user for `AddLiquidityInput` and `AddLiquidityKind`

test/v3/addLiquidityBoosted/addLiquidityBoosted.integration.test.ts
MattPereira: What do you mean by "wrapped WETH" ? is wwETH a new thing? 

Also friendly reminder we have [pool creation UI](https://pool-creator-git-v3-pool-creation-balancer.vercel.app/?_vercel_share=xpIdz0S1R08JOWZx6BhA38OLthCZyBTb) and [SDK script](https://github.com/balancer/b-sdk/blob/main/examples/createAndInitPool/createAndInitPoolV3.ts) for creating pools

test/v3/priceImpact/priceImpact.V3.integration.test.ts
```diff
@@ -28,7 +28,11 @@ const WETH = TOKENS[chainId].WETH;
 describe('PriceImpact V3', () => {
     let rpcUrl: string;
     beforeAll(async () => {
-        ({ rpcUrl } = await startFork(ANVIL_NETWORKS.SEPOLIA));
+        ({ rpcUrl } = await startFork(
+            ANVIL_NETWORKS.SEPOLIA,
+            undefined,
+            7178065n,
```
MattPereira, line 34: could the block number be imported from `anvil-global-setup` ?

MattPereira: err nvm it cant cus it was moved forward i see now

MattPereira: Ah i see mention in the PR comment now. I agree with this decision :+1:

test/v3/addLiquidityBoosted/addLiquidityBoosted.integration.test.ts
MattPereira: ah i see gotta init the buffer within the vault. i wonder if we will have UI for that eventually or if plan is to just create them internally as needed

MattPereira (approved):
Another good one! :trophy:

test/mockData/partialBoostedPool.ts
```diff
@@ -24,3 +24,28 @@ export const partialBoostedPool_USDT_stataDAI: PoolStateWithUnderlyings = {
...
+            underlyingToken: null,
+        },
+        {
+            index: 1,
+            address: '0x978206fae13faf5a8d293fb614326b237684b750',
+            decimals: 6,
+            underlyingToken: {
+                address: '0xaa8e23fb1079ea71e0a56f48a2aa51851d8433d0',
+                decimals: 6,
+                index: 1,
```
MattPereira, line 47: https://test.balancer.fi/pools/sepolia/v3/0x445A49D1Ad280B68026629fE029Ed0Fbef549a94
deploy12 :point_up:

brunoguerios:
A bit of context on the last commit I pushed:

It adapts the AddLiquidityProportional integration test to accept 1 wei diff when an amountIn is provided as referenceAmount
There is a scenario where you set the referenceAmount and it calculates a respective bptOut for it that requires rounding down.
This can cause the queryOutput to return an amountIn that is 1 wei below the user input.
There’s no way around it, so all I did was to relax the test to account for that.
Let me know if you think there’s a better approach

johngrantuk (approved):

PR #517, opened 2024-12-05
Update to deploy 12 :rocket: 

MattPereira:
### Notes
- Only updated addresses for `MAINNET` and `GNOSIS_CHAIN` to deploy12
  -  Left `SEPOLIA` on deploy11 while we wait for data & API to sync 
- Grabbed weighted and stable pool abis from `MockWeightedPool` and `MockStablePool` deployments 
- Not sure how to find `AUTHORIZER` addresses or how to be sure about `ADMIN_OF_AUTHORIZER` address
- Do we need to build `VaultExtension` abi from monorepo?
- I seem to have different prettier settings than whoever updated ABIs last :upside_down_face: 

### Contract Addresses
- https://github.com/balancer/balancer-deployments/pull/124
- https://github.com/balancer/balancer-deployments/pull/128
- https://github.com/balancer/balancer-deployments/pull/130
- https://github.com/balancer/balancer-deployments/pull/135


src/utils/constants.ts
```diff
@@ -150,11 +150,15 @@ export const VAULT: Record<number, Address> = {
 };
 
 export const VAULT_V3: Record<number, Address> = {
-    [ChainId.SEPOLIA]: '0xBC582d2628FcD404254a1e12CB714967Ce428915',
+    [ChainId.SEPOLIA]: '0xbA1333333333a1BA1108E8412f11850A5C319bA9',
```
johngrantuk, line 153: I don't think we can update SEPOLIA until we have the test pools, etc deployed which may take some time. If we need a release before then provided there are no breaking changes resulting from updated ABIs I think we can keep previous SEPOLIA and add MAINNET/GNOSIS.

MattPereira: that makes sense :ballot_box_with_check:

PR #522, opened 2024-12-07
chore: log multicall v2 pool state error

agualis:
We are hitting this error quite frequently from our integration tests in GHA. 

This PR adds a console.log to debug the issue.

### DIFF
--- a/.changeset/chilled-hornets-explode.md
+++ b/.changeset/chilled-hornets-explode.md
@@ -1,0 +1,5 @@
+---
+"@balancer/sdk": patch
+---
+
+chore: debug v2 pool state multicall errors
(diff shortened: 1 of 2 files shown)


brunoguerios (approved):
Sure! Let us know what you find, so we're able to fix on our end.

brunoguerios:
Do you happen to have the inputs that caused this error?
I can see it's adding liquidity proportionally to [this pool](https://balancer.fi/pools/polygon/v2/0xdac42eeb17758daa38caf9a3540c808247527ae3000200000000000000000a2b), but I can't reproduce the issue 😬 
I even added an integration test that is passing on my end.

agualis:
Thanks for checking Bruno 🙏
I will take a look at this after the launch

PR #530, opened 2024-12-11
Add support for InitBuffer

### LINKED ISSUE #519: Add support for Init Buffer

brunoguerios:
Close #519

### DIFF
--- a/src/entities/permit2Helper/index.ts
+++ b/src/entities/permit2Helper/index.ts
@@ -30,6 +30,7 @@ import { TokenAmount } from '../tokenAmount';
 import { permit2Abi } from '@/abi';
 import { getAmountsCall } from '../addLiquidity/helpers';
 import { AddLiquidityBufferBuildCallInput } from '../addLiquidityBuffer/types';
+import { InitBufferBuildCallInput } from '../initBuffer/types';
 
 export * from './allowanceTransfer';
 export * from './constants';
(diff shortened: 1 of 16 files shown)


src/entities/permit2Helper/index.ts
```diff
@@ -195,6 +196,46 @@ export class Permit2Helper {
...
+            await getDetails(
+                input.client,
+                input.underlyingAmountIn.token.address,
+                input.owner,
+                spender,
+                input.underlyingAmountIn.amount,
+                input.expirations ? input.expirations[1] : undefined,
+                input.nonces ? input.nonces[1] : undefined,
+            ),
+        ];
```
MattPereira, line 235: i would have tried to map this, but i like your approach better given the circumstances

test/v2/createPool/composableStable.integration.test.ts
```diff
@@ -18,14 +18,20 @@ import { ANVIL_NETWORKS, startFork } from '../../anvil/anvil-global-setup';
...
     let createPoolComposableStableInput: CreatePoolV2ComposableStableInput;
+    let rpcUrl: string;
+
     beforeAll(async () => {
+        // TODO: find out why findEventInReceiptLogs doesn't find the event when blockNumber is updated to 21373640n (this causes test to fail)
+        ({ rpcUrl } = await startFork(
+            ANVIL_NETWORKS.MAINNET,
+            undefined,
+            18980070n,
+        ));
```
MattPereira, line 34: `BalancerErrors.sol` looks like fun :gift: 
https://www.tdly.co/shared/simulation/85d1f5ff-5bd8-4bfb-9a20-d14eefebde6d

brunoguerios: Nice catch! 
The factory for CSPv5 has been disabled 👀 
Luiz was the one that was keeping track of factory updates and I missed this - I'll bump to CSPv6 over here.
And I'll assume we won't have a newer version on Balancer v2, so we should be ok with that.
But we'll need to keep that in mind to make sure we don't lag behind similar updates for Balancer v3 pools.

brunoguerios: I created an issue to keep track of this: https://github.com/balancer/b-sdk/issues/535

src/entities/permit2Helper/index.ts
brunoguerios: Yeah, I though of having an array of amountsIn and stating somewhere that index 0 was the wrapped token and index 1 was underlying token, but that felt a bit error prone.
Since it's always 2 tokens only, having them explicitly declared like this felt like a better approach.

test/v3/initBuffer/initBuffer.integration.test.ts
```diff
@@ -0,0 +1,230 @@
...
+                    rawAmount: parseUnits('100', USDC.decimals),
+                    decimals: USDC.decimals,
+                    address: USDC.address,
+                },
+                wrappedAmountIn: {
+                    rawAmount: parseUnits('0', gearboxUSDC.decimals),
+                    decimals: gearboxUSDC.decimals,
+                    address: gearboxUSDC.address,
+                },
+            };
```
MattPereira, line 115: You might consider moving this input declaration to global scope and assign the value in beforeAll since is the same for both w/ and w/o permit2

```
let initBufferInput: InitBufferInput;
...

        initBufferInput = {
            chainId,
            rpcUrl,
            underlyingAmountIn: {
                rawAmount: parseUnits('100', USDC.decimals),
                decimals: USDC.decimals,
                address: USDC.address,
            },
            wrappedAmountIn: {
                rawAmount: parseUnits('0', gearboxUSDC.decimals),
                decimals: gearboxUSDC.decimals,
                address: gearboxUSDC.address,
            },
        };
```

MattPereira: Also just to make sure, the reason you're setting wrappedAmountIn to 0 is because would be a pain to do the token slot thingy to set test user balance for gearbox? 

But if user did have wrapped amount they could use them as part of init right?

MattPereira (approved):
Looks good friend! :muscle:

test/v3/initBuffer/initBuffer.integration.test.ts
brunoguerios: Yeah, any amount of wrapped/underlying can be provided to init the buffer. I set it to zero to avoid the setup pain 😅 
You could think that since it's zero, we could set the token to undefined, but we need the address to be provided.

test/v2/createPool/composableStable.integration.test.ts
johngrantuk: Does that mean we can remove the TODO?

johngrantuk (approved):

PR #533, opened 2024-12-11
Remove duplicate queryRemoveLiquidityRecovery

### LINKED ISSUE #531: Remove duplicate `queryRemoveLiquidityRecovery` for v3
This method was implemented when we didn't have query available for removeLiquidityRecovery, but now we do and it's already being handled inside the default Remove Liquidity `query`, so we can remove it as duplicate.

Note: v2 would still need the custom implementation of query for remove liquidity recovery, as there's no actual query available on vault (v2). But in any case, we should consider a refactor where `queryRemoveLiquidityRecovery` custom implementation is moved to inside `query` to make sure both v2 and v3 share the same external interface.

brunoguerios:
Closes #531 

-- IMPORTANT --
Please note that this bumps a `major version` as it's a breaking change to the external interface.
I think that should be ok as we're preparing to release v1.0

### DIFF
--- a/.changeset/silly-dolls-matter.md
+++ b/.changeset/silly-dolls-matter.md
@@ -1,0 +1,5 @@
+---
+"@balancer/sdk": major
+---
+
+Remove duplicate queryRemoveLiquidityRecovery
--- a/.changeset/silly-dolls-matter.md
+++ b/.changeset/silly-dolls-matter.md
@@ -2,4 +2,4 @@
 "@balancer/sdk": major
 ---
 
-Remove duplicate queryRemoveLiquidityRecovery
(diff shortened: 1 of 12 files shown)


MattPereira (approved):
Good idea and squeaky clean! :soap:

johngrantuk (approved):
Nice to tidy.
I think it might be helpful to extend the changset a bit to explain whats changed and how a user can update on their side.

johngrantuk:
> @johngrantuk - let me know what you think about the new changeset message:
> 
> > Remove `queryRemoveLiquidityRecovery` from `RemoveLiquidity`. Please use `query` with `RemoveLiquidityRecoveryInput` instead.

Looks good, thanks.

PR #547, opened 2024-12-18
Fix Composable Stable CreatePool v2 integration test

### LINKED ISSUE #548: Fix composable stable create pool v2 flaky test
It's failing intermittently on CI

Note: likely related to us relying on finding event on receipt logs to assert the pool was properly created? Can we leverage the `create` function output? All different pool factories return the same output for the create function, which is the pool address.

agualis:
Closes #548


brunoguerios:
As already mentioned in #548 - we won't re-enable these tests for now.

PR #553, opened 2025-01-10
Add permit2 for init pool

### LINKED ISSUE #476: Add permit2 support for pool initialization
We want to make it easier for SDK consumers init pools using permit2

### Items
- [x] Add a `signInitPoolApproval` function to [permit2helper](https://github.com/balancer/b-sdk/blob/main/src/entities/permit2Helper/index.ts)
- [x] Add `.buildCallWithPermit2` to `InitPoolV3`
- [x] Update create v3 stable pool test 
- [x] Update create v3 weighted pool test
- [x] Update v3 pool creation example 
- [ ] After merge, update v3 pool creation UI to use it

MattPereira:
Closes #476 

### Summary
- For testing, we could modify [doInitPool](https://github.com/balancer/b-sdk/blob/74d61ab3476134b2406bc14f33e81b8e83752798/test/lib/utils/initPoolHelper.ts#L23-L46)  to handle both v2 and v3, but I'm proposing we favor the more straightforward approach
- Refactored pool creation example to offer functions for both fork and live network
- Refactored `getDetails` of `Permit2Helper` to allow for type `Account` so that `client.signTypedData` can be run with a script

### DIFF
--- a/examples/createAndInitPool/createAndInitPoolV3.ts
+++ b/examples/createAndInitPool/createAndInitPoolV3.ts
@@ -36,6 +36,7 @@ import {
     PERMIT2,
     erc20Abi,
     Permit2Helper,
+    PublicWalletClient,
 } from 'src';
 import { startFork, ANVIL_NETWORKS } from 'test/anvil/anvil-global-setup';
 import { findEventInReceiptLogs } from 'test/lib/utils/findEventInReceiptLogs';
@@ -46,6 +47,7 @@ import { approveSpenderOnTokens } from 'test/lib/utils/helper';
 // Create pool config
 const chainId = ChainId.SEPOLIA;
 const poolType = PoolType.Stable;
+const protocolVersion = 3;
 const swapFeePercentageDecimals = 16;
 const AAVE_FAUCET_USDT = {
     address: '0xaa8e23fb1079ea71e0a56f48a2aa51851d8433d0' as `0x${string}`,
@@ -81,7 +83,7 @@ const createPoolInput: CreatePoolV3StableInput = {
     poolHooksContract: zeroAddress,
     enableDonation: false,
     disableUnbalancedLiquidity: false,
-    protocolVersion: 3,
+    protocolVersion,
     chainId,
 };
 
@@ -89,12 +91,12 @@ const createPoolInput: CreatePoolV3StableInput = {
 const amountsIn = [
     {
         address: AAVE_FAUCET_USDT.address,
-        rawAmount: parseUnits('3', AAVE_FAUCET_USDT.decimals),
+        rawAmount: parseUnits('1', AAVE_FAUCET_USDT.decimals),
         decimals: AAVE_FAUCET_USDT.decimals,
     },
     {
         address: STATA_ETH_DAI.address,
-        rawAmount: parseUnits('3', STATA_ETH_DAI.decimals),
+        rawAmount: parseUnits('1', STATA_ETH_DAI.decimals),
         decimals: STATA_ETH_DAI.decimals,
     },
 ];
@@ -114,15 +116,15 @@ export async function runAgainstFork() {
     })
         .extend(publicActions)
         .extend(walletActions);
-    const userAccount = (await client.getAddresses())[0];
+    const account = (await client.getAddresses())[0];
 
     // Create the pool and parse the tx receipt to get the pool address
-    const poolAddress = await createPool({ client, userAccount });
+    const poolAddress = await createPool({ client, account });
 
     // Approve the cannonical permit2 contract to spend tokens that will be used for pool init
     await approveSpenderOnTokens(
         client,
-        userAccount,
+        account,
         amountsIn.map((t) => t.address),
         PERMIT2[chainId],
     );
@@ -131,7 +133,7 @@ export async function runAgainstFork() {
     const initPoolCall = await buildInitPoolCall({
         client,
         rpcUrl,
-        userAccount,
+        account,
         poolAddress,
     });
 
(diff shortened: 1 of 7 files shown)


src/entities/permit2Helper/index.ts
```diff
@@ -358,7 +358,20 @@ const getNonce = async (
         functionName: 'allowance',
         args: [owner, token, spender],
     });
-    const nonce = result[1];
+    const nonce = result[2];
```
MattPereira, line 361: I think nonce is the third value returned by `allowance`
https://github.com/Uniswap/permit2/blob/cc56ad0f3439c502c246fc5cfcc3db92bb8b7219/src/interfaces/IAllowanceTransfer.sol#L108-L114

examples/createAndInitPool/createAndInitPoolV3.ts
```diff
@@ -58,83 +58,122 @@ const STATA_ETH_DAI = {
...
+                address: a.address,
+                slot: getSlot(chainId, a.address),
+                rawBalance: a.rawAmount,
+            })),
         },
-    ];
+        [...amountsIn.map((a) => a.address), poolAddress],
+        createPoolInput.protocolVersion,
+    );
+}
```
MattPereira, line 156: `runAgainstFork` reverts on pool init, but I'm having a tough time debugging why

examples/createAndInitPool/createAndInitPoolV3.ts
```diff
@@ -244,71 +228,45 @@ async function createPool({ client, call, userAccount }) {
...
-        details,
-        spender: BALANCER_ROUTER[chainId],
-        sigDeadline: MaxSigDeadline,
-    };
+    // Sign permit2 approval
+    const permit2 = await Permit2Helper.signInitPoolApproval({
+        ...initPoolInput,
+        client,
+        owner: account,
+    });
```
MattPereira, line 260: `runAgainstNetwork` fails here because of an unsupported method

```
HttpRequestError: HTTP request failed.

Status: 400

Details: {"code":-32600,"message":"Unsupported method: eth_signTypedData_v4. }
```

MattPereira: I confirmed that it is possible to modify the type for [owner](https://github.com/balancer/b-sdk/blob/74d61ab3476134b2406bc14f33e81b8e83752798/src/entities/permit2Helper/index.ts#L293) of `Permit2Helper.signPermit2` to be `Address | Account`  and that makes the script work if I provide an `Account`, but I'm assuming I am missing something about the client setup because I saw the v3 frontend repo is able to pass just an address and the `Permit2Helper` works there :thinking:

MattPereira: Here is how I'm creating the client for reference
https://github.com/balancer/b-sdk/blob/c7a6c0d62fc064cd6dcb1a7054c9550cec4dd02b/examples/createAndInitPool/createAndInitPoolV3.ts#L162-L168

examples/createAndInitPool/createAndInitPoolV3.ts
```diff
@@ -29,112 +32,142 @@ import {
...
 
-    // Submit the create pool tx and wait for the PoolCreated event to find pool address
-    const poolAddress = await createPool({ client, call, userAccount });
-    console.log('Created Pool Address: ', poolAddress);
+    const client = createWalletClient({
+        chain: CHAINS[chainId],
+        transport: http(rpcUrl),
+        account: privateKeyToAccount(process.env.PRIVATE_KEY as `0x${string}`),
+    })
+        .extend(walletActions)
```
johngrantuk, line 161: I don't think this is required as its already a wallet client?

src/entities/permit2Helper/index.ts
johngrantuk: Looks like you're right! Not sure why this hasn't been an issue until now.

examples/createAndInitPool/createAndInitPoolV3.ts
johngrantuk: Everything looks right on your client setup from what I can see.
Tracing into the viem code it seems like an Address is parsed to a `type: 'json-rpc'` which then branches to a client request:
```
client.request({
        method: 'eth_signTypedData_v4',
        params: [account.address, typedData],
    }
```
So it appears we have to use an `Account`

src/entities/permit2Helper/index.ts
MattPereira: I think the v3 FE is fetching the nonce manually instead of letting the SDK grab it

examples/createAndInitPool/createAndInitPoolV3.ts
MattPereira: you are correct ser!

src/entities/permit2Helper/index.ts
```diff
@@ -323,15 +323,16 @@ const signPermit2 = async (
...
     spender: Address,
     amount = MaxAllowanceTransferAmount,
     expiration = Number(MaxAllowanceExpiration),
     nonce?: number,
 ) => {
+    const _owner = typeof owner === 'string' ? owner : owner.address;
     let _nonce: number;
     if (nonce === undefined) {
-        _nonce = await getNonce(client, token, owner, spender);
+        _nonce = await getNonce(client, token, _owner, spender);
```
MattPereira, line 335: @brunoguerios lmk if you agree with this approach and if we should handle the rest of the sign helper methods as part of the PR or open a new issue

brunoguerios: Sure! It does seem a more robust approach. If it's just a matter of adding that single line for `owner` type check on each method, I'd go ahead and make the change 👍

brunoguerios:
No idea why unrelated swap tests are failing 🤔 
Are you able to reproduce locally?
If not, I'd simply re-run tests to see if they pass.

brunoguerios (approved):
LGTM ✅ 
Ps: thanks for fixing/refactoring a bunch of the Permit code - it does look a lot better 👏

PR #572, opened 2025-01-25
New composite router for boosted pool operations

### LINKED ISSUE #555: Update to new CompositeLiquidity and Nested Routers
!!! Currently blocked by SC
* Composite adds useWrapped[] bool
* NestedPools will also be updated but interface shouldn't change, just needs updated address (and tests should pass)

MattPereira:
Closes #555 

### Summary
- Temporarily splitting composite router into boosted and nested versions
  - See why [here](https://balancerecosystem.slack.com/archives/C08AG424F09/p1737697686330499)
- New composite router allows for passing array of booleans that determine if token should be wrapped / unwrapped

### Update 1
- For proportional boosted queries, using the new `tokensIn` and `tokensOut` that are returned by composite router 

### Update 2
- Major refactor of implementation to infer `wrapUnderlying` and `unwrapWrapped` so user only has to worry about which tokens they want to add / remove

### DIFF
--- a/.changeset/red-buttons-check.md
+++ b/.changeset/red-buttons-check.md
@@ -1,5 +1,5 @@
 ---
-"@balancer/sdk": minor
+"@balancer/sdk": major
 ---
 
 new composite router for boosted operations with ability to wrap/unwrap single token
--- a/.changeset/red-buttons-check.md
+++ b/.changeset/red-buttons-check.md
@@ -1,0 +1,5 @@
+---
+"@balancer/sdk": minor
+---
+
+new composite router for boosted operations with ability to wrap/unwrap single token
--- a/src/entities/addLiquidityBoosted/index.ts
+++ b/src/entities/addLiquidityBoosted/index.ts
@@ -55,9 +55,7 @@ export class AddLiquidityBoostedV3 {
         });
 
         const bptToken = new Token(input.chainId, poolState.address, 18);
-        const wrapUnderlying: boolean[] = new Array(
-            poolState.tokens.length,
-        ).fill(false);
+        const wrapUnderlying: boolean[] = [];
 
         let bptOut: TokenAmount;
         let amountsIn: TokenAmount[];
@@ -89,8 +87,8 @@ export class AddLiquidityBoostedV3 {
         switch (input.kind) {
             case AddLiquidityKind.Unbalanced: {
                 // Infer wrapUnderlying from token addreses provided by amountsIn
-                const tokensIn: ExtendedMinimalToken[] = input.amountsIn.map(
-                    (amountIn) => {
+                const tokensIn: ExtendedMinimalToken[] = input.amountsIn
+                    .map((amountIn) => {
                         const amountInAddress = amountIn.address.toLowerCase();
                         const token = poolStateTokenMap[amountInAddress];
                         if (!token) {
@@ -99,10 +97,11 @@ export class AddLiquidityBoostedV3 {
                             );
                         }
                         return token;
-                    },
-                );
+                    })
+                    .sort((a, b) => a.index - b.index); // sort by index so wrapUnderlying is in correct order
+
                 tokensIn.forEach((t) => {
-                    wrapUnderlying[t.index] = t.isUnderlyingToken;
+                    wrapUnderlying.push(t.isUnderlyingToken);
                 });
 
                 // It is allowed not not provide the same amount of TokenAmounts as inputs
@@ -131,14 +130,19 @@ export class AddLiquidityBoostedV3 {
             case AddLiquidityKind.Proportional: {
                 // User provides tokensIn addresses so we can infer if they need to be wrapped
 
-                input.tokensIn.forEach((t) => {
-                    const tokenWithDetails =
-                        poolStateTokenMap[t.toLowerCase() as Address];
-                    if (!tokenWithDetails) {
-                        throw new Error(`Invalid token address: ${t}`);
-                    }
-                    wrapUnderlying[tokenWithDetails.index] =
-                        tokenWithDetails.isUnderlyingToken;
+                const sortedTokensInWithDetails = input.tokensIn
+                    .map((t) => {
+                        const tokenWithDetails =
+                            poolStateTokenMap[t.toLowerCase() as Address];
+                        if (!tokenWithDetails) {
+                            throw new Error(`Invalid token address: ${t}`);
+                        }
+                        return tokenWithDetails;
+                    })
+                    .sort((a, b) => a.index - b.index);
+
+                sortedTokensInWithDetails.forEach((t) => {
+                    wrapUnderlying.push(t.isUnderlyingToken);
                 });
 
                 const bptAmount = await getBptAmountFromReferenceAmountBoosted(
--- a/src/entities/addLiquidityBoosted/index.ts
+++ b/src/entities/addLiquidityBoosted/index.ts
@@ -55,10 +55,12 @@ export class AddLiquidityBoostedV3 {
         });
 
         const bptToken = new Token(input.chainId, poolState.address, 18);
+        const wrapUnderlying: boolean[] = new Array(
+            poolState.tokens.length,
+        ).fill(false);
 
         let bptOut: TokenAmount;
         let amountsIn: TokenAmount[];
-        let wrapUnderlying: boolean[];
 
         type ExtendedMinimalToken = MinimalToken & {
             isUnderlyingToken: boolean;
@@ -87,8 +89,8 @@ export class AddLiquidityBoostedV3 {
         switch (input.kind) {
             case AddLiquidityKind.Unbalanced: {
                 // Infer wrapUnderlying from token addreses provided by amountsIn
-                const tokensIn: ExtendedMinimalToken[] = input.amountsIn
-                    .map((amountIn) => {
+                const tokensIn: ExtendedMinimalToken[] = input.amountsIn.map(
+                    (amountIn) => {
                         const amountInAddress = amountIn.address.toLowerCase();
                         const token = poolStateTokenMap[amountInAddress];
                         if (!token) {
@@ -96,17 +98,18 @@ export class AddLiquidityBoostedV3 {
                                 `Token not found in poolState: ${amountInAddress}`,
                             );
                         }
-
                         return token;
-                    })
-                    .sort((a, b) => a.index - b.index); // Sort by index to ensure match for wrapUnderlying values
+                    },
+                );
+                tokensIn.forEach((t) => {
+                    wrapUnderlying[t.index] = t.isUnderlyingToken;
+                });
 
                 // It is allowed not not provide the same amount of TokenAmounts as inputs
                 // as the pool has tokens, in this case, the input tokens are filled with
                 // a default value ( 0 in this case ) to assure correct amounts in as the pool has tokens.
                 const sortedTokens = getSortedTokens(tokensIn, input.chainId);
                 const maxAmountsIn = getAmounts(sortedTokens, input.amountsIn);
-                wrapUnderlying = tokensIn.map((t) => t.isUnderlyingToken);
 
                 const bptAmountOut = await doAddLiquidityUnbalancedQuery(
                     input.rpcUrl,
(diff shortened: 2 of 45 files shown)


src/entities/priceImpact/addLiquidityNested.ts
```diff
@@ -134,6 +134,7 @@ async function getAddBoostedUnbalancedResult(
         rpcUrl,
         amountsIn,
         kind: AddLiquidityKind.Unbalanced,
+        wrapUnderlying: [true, true],
```
MattPereira, line 137: This is bad, but `wrapUnderlying` has to be there for price impact test to pass. Not sure how to handle cus adding `wrapUnderlying` to `AddLiquidityNestedInput` type would blow up nested which is not supported by the new router

test/v3/removeLiquidityBoosted/removeLiquidityBoosted.integration.test.ts
```diff
@@ -133,11 +121,16 @@ describe('remove liquidity test', () => {
...
             kind: AddLiquidityKind.Proportional,
+            wrapUnderlying: [true, true],
         };
 
-        const _addLiquidityOutput = await doAddLiquidity({
-            ...txInput,
-            addLiquidityInput: input,
+        await doAddLiquidity({
+            client,
+            addLiquidity: new AddLiquidityBoostedV3(), // TODO: debug linter type complaint
```
MattPereira, line 129: Not sure best way to handle this linter complaint that predates this PR

src/entities/addLiquidityBoosted/index.ts
```diff
@@ -26,21 +26,21 @@ import {
...
     AddLiquidityBoostedBuildCallInput,
     AddLiquidityBoostedInput,
     AddLiquidityBoostedQueryOutput,
 } from './types';
+import { AddLiquidity } from '../addLiquidity';
 
-export class AddLiquidityBoostedV3 {
-    private readonly inputValidator: InputValidator = new InputValidator();
-
+export class AddLiquidityBoostedV3 extends AddLiquidity {
```
MattPereira, line 43: Was running into type errors trying to do this in a test

```
        await doAddLiquidity({
            client,
            addLiquidity: new AddLiquidityBoostedV3(),
```

I think is better to extend `AddLiquidity` class instead of creating a new InputValidator here, but it requires changing inputValidator from `private` to `protected` within the `AddLiquidty` class

johngrantuk: Not sure what the benefits are of extending? Aren't we just over writing all the methods?

src/entities/priceImpact/addLiquidityNested.ts
```diff
@@ -134,6 +134,7 @@ async function getAddBoostedUnbalancedResult(
         rpcUrl,
         amountsIn,
         kind: AddLiquidityKind.Unbalanced,
+        wrapUnderlying: [true, true], // TODO: Handle more elegantly?
```
johngrantuk, line 137: Yeah, should probably handle correctly if we are going to support this. Can you figure out an example of what this looks like and if wrap underlying is valid - I think its a nested boosted but haven't thought it through properly.

test/v3/addLiquidityBoosted/addLiquidityBoosted.integration.test.ts
```diff
@@ -39,11 +40,12 @@ import {
 import { ANVIL_NETWORKS, startFork } from '../../anvil/anvil-global-setup';
 import { boostedPool_USDC_USDT } from 'test/mockData/boostedPool';
 
-const protocolVersion = 3;
+// const protocolVersion = 3;
```
johngrantuk, line 43: just delete it

johngrantuk (changes requested):
Just a few loose ends to tidy and a changeset to add.

test/v3/addLiquidityBoosted/addLiquidityBoosted.integration.test.ts
```diff
@@ -53,7 +55,11 @@ describe('Boosted AddLiquidity', () => {
     const addLiquidityBoosted = new AddLiquidityBoostedV3();
 
     beforeAll(async () => {
-        ({ rpcUrl } = await startFork(ANVIL_NETWORKS[ChainId[chainId]]));
+        ({ rpcUrl } = await startFork(
+            ANVIL_NETWORKS[ChainId[chainId]],
+            undefined,
+            7562540n, // block after new composite liquidity router deployed
```
brunoguerios, line 61: Did you try bumping the global blockNumber instead? Tests are more performant if we have all running on the same block.
If we still have other tests that are block dependent, then maybe hardcode those with the previous block and add an issue to try to improve them later.

src/entities/addLiquidityBoosted/index.ts
MattPereira: @johngrantuk there was a previously existing type error in the test

image

To fix it I'm proposing the extend but I dont feel strongly about it. Could rewrite the test to not use `doAddLiquidity` or we could just be okay with the type error. let me know

You should be able to see type error on existing main branch [here](https://github.com/balancer/b-sdk/blob/3554d0cb2dee7450c29f014269778c750ddcdd26/test/v3/removeLiquidityBoosted/removeLiquidityBoosted.integration.test.ts#L120)

johngrantuk: Can potentially fix by doing something like below [here](https://github.com/balancer/b-sdk/blob/88b755b1db04da7fd77c2df41c4e37a56ebd8dca/test/lib/utils/types.ts#L21)?:
```
export type AddLiquidityTxInput = {
    client: PublicWalletClient & TestActions;
    addLiquidity: AddLiquidity | AddLiquidityBoostedV3;
    addLiquidityInput: AddLiquidityInput | AddLiquidityBoostedInput;
    slippage: Slippage;
    poolState: PoolState;
    testAddress: Address;
    wethIsEth?: boolean;
    fromInternalBalance?: boolean;
    usePermit2Signatures?: boolean;
};
```

MattPereira: ya that works. change is made ✅

brunoguerios (changes requested):
Can you please add an add liquidity proportional test covering a scenario that was not previously supported?
I believe that will expose an issue with existing calculateProportionalAmounts helper not being able to handle these new possibilities for amountsIn combinations.

src/entities/priceImpact/addLiquidityNested.ts
MattPereira: I think this may be a decent way to handle  ( until composite router v3 with support for both boosted and nested )

https://github.com/balancer/b-sdk/blob/12d5255f94dce0268ce9b18b3be1ea836ec28117/src/entities/priceImpact/addLiquidityNested.ts#L67

MattPereira:
@brunoguerios added tests for add with single wrap and remove with single unwrap. Also, thanks again for the assist! :heart:

src/entities/inputValidator/boosted/inputValidatorBoosted.ts
```diff
@@ -14,20 +14,22 @@ export class InputValidatorBoosted extends InputValidatorBase {
...
-                    return t.underlyingToken.address.toLowerCase();
-                return t.address.toLowerCase();
-            });
+            // List of all tokens that can be added to the pool
+            const poolTokens = poolState.tokens
+                .flatMap((token) => [
+                    token.address.toLowerCase(),
+                    token.underlyingToken?.address.toLowerCase(),
+                ])
+                .filter(Boolean);
```
johngrantuk, line 23: Never seen this before - learn something new!

src/entities/removeLiquidityBoosted/doRemoveLiquidityProportionalQuery.ts
```diff
@@ -10,20 +19,28 @@ export const doRemoveLiquidityProportionalQuery = async (
     sender: Address,
     userData: Hex,
     poolAddress: Address,
+    unwrapWrapped: boolean[],
     block?: bigint,
-): Promise<bigint[]> => {
+): Promise<[`0x${string}`[], bigint[]]> => {
```
johngrantuk, line 24: Can this just be `Address`?

.changeset/red-buttons-check.md
johngrantuk, line 1: I believe this should be a major version bump because: 
`MAJOR version when you make incompatible API changes` ([see](https://semver.org/))

johngrantuk (approved):

src/entities/addLiquidityBoosted/index.ts
```diff
@@ -85,6 +87,7 @@ export class AddLiquidityBoostedV3 {
                     input.sender ?? zeroAddress,
                     input.userData ?? '0x',
                     poolState.address,
+                    input.wrapUnderlying,
                     maxAmountsIn,
```
brunoguerios, line 91: I believe this `maxAmountsIn` is relying on `childTokens`, which needs to be updated here for add liquidity unbalanced (similar to what you did on the add liquidity proportional case)

.changeset/red-buttons-check.md
MattPereira: good call that makes sense for warning consumers of danger

src/entities/addLiquidityBoosted/index.ts
MattPereira: nice catch! :raised_hands:

src/entities/addLiquidityBoosted/index.ts
```diff
@@ -56,27 +58,43 @@ export class AddLiquidityBoostedV3 {
...
-                address: t.address,
-                decimals: t.decimals,
-                index: t.index,
-            };
-        });
-
         switch (input.kind) {
             case AddLiquidityKind.Unbalanced: {
+                // Use poolState to add token index to amountsIn
+                const tokensIn = input.amountsIn.map((amountIn) => {
```
brunoguerios, line 64: It feels like a lot of this `tokensIn` logic is repeated within unbalanced and proportional implementations. Do you think it's worth extracting/reusing?

brunoguerios: given we're in a bit of a tight schedule, we can leave this refactor to be done in a new (low-priority) PR

MattPereira: yup I agree on both notes :+1:

src/entities/priceImpact/addLiquidityNested.ts
MattPereira: turns out this was just a symptom of bad initial implementation plan :sweat:

src/entities/addLiquidityBoosted/index.ts
```diff
@@ -128,26 +148,13 @@ export class AddLiquidityBoostedV3 {
...
-                        );
-                    })?.decimals;
-
-                    if (!decimals)
-                        throw new Error(
-                            `Token decimals missing for ${tokenInAddress}`,
-                        );
-
+                amountsIn = tokensIn.map((t, i) => {
+                    const tokenInAddress = t.toLowerCase() as `0x${string}`;
```
brunoguerios, line 156:
```suggestion
                    const tokenInAddress = t.toLowerCase() as Address;
```

src/entities/priceImpact/addLiquidityUnbalancedBoosted.ts
```diff
@@ -66,7 +66,7 @@ export async function addLiquidityUnbalancedBoosted(
         chainId: input.chainId,
         rpcUrl: input.rpcUrl,
         bptIn: bptOut.toInputAmount(),
-        unwrapWrapped: input.wrapUnderlying,
+        tokensOut: poolTokens.map((t) => t.address),
```
brunoguerios, line 69: I'll double check over here, but I'd expect these to be `amountsIn` tokens

src/entities/utils/boostedPoolHelpers.ts
```diff
@@ -0,0 +1,65 @@
...
+ *   },
+ *   "0xaa8e23fb1079ea71e0a56f48a2aa51851d8433d0": {
+ *     index: 1,
+ *     decimals: 6,
+ *     address: "0xaa8e23fb1079ea71e0a56f48a2aa51851d8433d0",
+ *     isUnderlyingToken: true
+ *   }
+ * }
+ */
+export function buildPoolStateTokenMap(
```
brunoguerios, line 41: Nice helper 👏

brunoguerios (approved):
I like this solution a lot better 😃 👏 
Thanks for following up on the requested changes!
There is one detail on the priceImpact side of things that I'd like to double check, but other than that it LGTM! ✅

src/entities/priceImpact/addLiquidityUnbalancedBoosted.ts
brunoguerios: yeah - we'll have to fix this one - I'll take a look over here

brunoguerios: ops - my bad - I didn't see that poolTokens are actually being mapped from amountsIn - this is correct ✅

PR #580, opened 2025-02-04
feat: Add create support for StableSurge (Sepolia address only).

johngrantuk:


johngrantuk:
Both failing tests are passing locally. Not sure what the issue is :(

src/entities/createPool/types.ts
```diff
@@ -59,6 +59,14 @@ export type CreatePoolV3StableInput = CreatePoolV3BaseInput & {
         paysYieldFees: boolean;
     }[];
 };
+
+export type CreatePoolStableSurgeInput = Omit<
+    CreatePoolV3StableInput,
+    'poolHooksContract' | 'poolType'
```
mkflow27, line 65: Omiting the poolHooksContract since there is a canonical BLabs one, I assume?

mkflow27 (approved):

brunoguerios:
> Both failing tests are passing locally. Not sure what the issue is :(

Are they timing out? Locally I had to bump `hookTimeout` to `120_000` to make it work.

test/v3/createPool/stableSurge/stableSurge.integration.test.ts
```diff
@@ -0,0 +1,179 @@
...
+                        index: 0,
+                    },
+                    {
+                        ...scDAI,
+                        index: 1,
+                    },
+                ],
+            },
+            permit2,
+        );
```
MattPereira, line 166: Curious about the decision to avoid using the poolState helper here. Fewer reads to optimize the speed of sluggish test?

https://github.com/balancer/b-sdk/blob/4f07d199b78720816c63743efd6afae67aa7f21f/test/v3/createPool/stable/stable.integration.test.ts#L110-L116

https://github.com/balancer/b-sdk/blob/4f07d199b78720816c63743efd6afae67aa7f21f/test/v3/createPool/stable/stable.integration.test.ts#L156-L160

src/entities/createPool/types.ts
MattPereira: Looks like the new `StableSurgeFactory` couples the factory's deployment with the hook :eyes:

https://github.com/balancer/balancer-v3-monorepo/blob/4680d59ede505847163197885780169af8873ec1/pkg/pool-hooks/contracts/StableSurgePoolFactory.sol#L27-L39

MattPereira: Perhaps we might consider changing `poolHooksContract` to be an optional property on the `CreatePoolV3BaseInput` type and then doing like 

```
input.poolHooksContract ?? zeroAddress,
```

for the createPoolStable and Weighted?

https://github.com/balancer/b-sdk/blob/4f07d199b78720816c63743efd6afae67aa7f21f/src/entities/createPool/createPoolV3/stable/createPoolStable.ts#L45

MattPereira (approved):
This approach should work well for the pool creator :pray:

johngrantuk:
> > Both failing tests are passing locally. Not sure what the issue is :(
> 
> Are they timing out? Locally I had to bump `hookTimeout` to `120_000` to make it work.

There was a bunch of weirdness going on. Was sporadically failing on random tests that were not associated with the PR. Was happening locally and on CI. Then it seemed to go away. Really not sure what was happening.

src/entities/createPool/types.ts
johngrantuk: With the goal of making it easier for the user or something else?

MattPereira: yeah would abstract away the need for user to know that they have to pass the zero address if they don't want to use a hook when creating pools with the core stable and weighted factories

Also to prevent the need for adding `<Omit ..., 'poolHooksContract` each time a new pool factory tied to a hook is added

PR #581, opened 2025-02-04
fix: wrong composable stable

johngrantuk:


test/validateNestedPoolState.test.ts
```diff
@@ -244,3 +260,66 @@ const deepPoolState: NestedPoolState = {
         },
     ],
 };
+
+const nestedOPoolState = {
```
agualis, line 264: @brunoguerios I just created a quick test but naming should be adapted to better express the intention of the test. Could you improve it please?

PR #584, opened 2025-02-05
Fix price impact for add liquidity boosted on near proportional inputs

brunoguerios:


brunoguerios:
Tests are not failing locally - any idea what's causing this unrelated test to fail on CI? 👀

examples/priceImpact/addLiquidityBoosted.ts
MattPereira, line 1: I could be using the wrong rpc url but this example fails on my machine

```
Error: addLiquidity operation will fail at SC level with user defined input.
ContractFunctionExecutionError: Missing or invalid parameters.
Double check you have provided the correct parameters.

URL: https://rpc.ankr.com/sonic_mainnet
Request body: {"method":"eth_call","params":[{"data":"0xc9b8ddb200000000000000000000000043026d483f42fb35efe03c20b251142d022783f200000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000160000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000186a000000000000000000000000000000000000000000000000000000000000186a00000000000000000000000000000000000000000000000000000000000000000"},"latest"]}
 
Raw Call Arguments:
  data:  0xc9b8ddb200000000000000000000000043026d483f42fb35efe03c20b251142d022783f200000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000160000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000186a000000000000000000000000000000000000000000000000000000000000186a00000000000000000000000000000000000000000000000000000000000000000
 
Contract Call:
  function:  queryAddLiquidityUnbalancedToERC4626Pool(address pool, bool[] wrapUnderlying, uint256[] exactAmountsIn, address sender, bytes userData)
  args:                                              (0x43026d483f42fb35efe03c20b251142d022783f2, [false,false], ["100000","100000"], 0x0000000000000000000000000000000000000000, 0x)
```

brunoguerios: true - I initially branched out v1.4.3 - it was working there 🫠 
I know that `buildPoolStateTokenMap` should be fixed to use `t.index` instead of `underlying.index` but that alone isn't enough to get the example to work
I'll take another look tomorrow to investigate this

brunoguerios: my bad - this won't work until the new composite router is deployed on Sonic 🤦‍♂️ - this fix was meant to be added as a hotfix to 1.4.3, so we'll have to put this on hold until everything else is deployed/updated

johngrantuk (approved):

PR #585, opened 2025-02-07
split into constants,v2,v3

mkflow27:
This pr closes https://github.com/orgs/balancer/projects/9/?pane=issue&itemId=94402497&issue=balancer%7Cb-sdk%7C563 The constants were moved either to new file (v2, or v3) or left in the old one


brunoguerios:
I like the idea of splitting the constants into 2 files so they are more organized, but if possible I'd like for constants themselves to also be differentiated with V2/V3 in their names.
I mean, reading the code is much easier to spot something like `VAULT_V2` and `VAULT_V3` than simply `VAULT` and having to look at the import to see it was imported from `constantsV2` or `constantsV3` file.
What do you think?

mkflow27 (commented):
> I like the idea of splitting the constants into 2 files so they are more organized, but if possible I'd like for constants themselves to also be differentiated with V2/V3 in their names.
> I mean, reading the code is much easier to spot something like `VAULT_V2` and `VAULT_V3` than simply `VAULT` and having to look at the import to see it was imported from `constantsV2` or `constantsV3` file.
> What do you think? 


I think it is a good suggestion which will make the code more readable. I will implement it. 
Edit: After looking through the code again, the only variable with a unclear naming would be vault -> Vault_v2. All other names can be tied to either general, v2 or v2 based on their name.

test/lib/utils/addLiquidityHelper.ts
```diff
@@ -9,19 +9,19 @@ import {
...
     TokenAmount,
-    VAULT,
     Permit2Helper,
     AddLiquidityKind,
     ChainId,
     isSameAddress,
     PublicWalletClient,
 } from 'src';
+import { VAULT_V2 } from '@/utils/constantsV2';
+import { BALANCER_ROUTER } from '@/utils/constantsV3';
```
brunoguerios, line 24: I know this might be a bit annoying, but can you please export both constants files on `utils/index.ts` and import them from `src` here (and in other test files)?
This way, we make sure those constants are exposed externally.

src/entities/utils/getPoolStateWithBalancesV2.ts
```diff
@@ -7,7 +7,8 @@ import {
 } from 'viem';
 import { PoolType } from '@/types';
 import { HumanAmount } from '@/data';
-import { CHAINS, VAULT } from '@/utils';
+import { CHAINS } from '@/utils';
+import { VAULT_V2 } from '@/utils/constantsV2';
```
brunoguerios, line 11: All these import changes could be avoided by applying the suggestion I made on the other comment -> export constantsV2 and constantsV3 files within utils/index.

brunoguerios (changes requested):

brunoguerios (approved):
Thanks for applying the requested changes! LGTM ✅

PR #590, opened 2025-02-10
Export stable surge factory abi

MattPereira:
in general i think exporting all abis from SDK is nice so devs can easily import source of truth for parsing event logs


johngrantuk (approved):
Makes sense, thanks!

PR #594, opened 2025-02-13
Add underlying tokens to poolState query

MattPereira:
Presently, I think user has to fetch the data from our API on their own if they want to use an underlying token to add liquidity to a boosted pool

( The boosted tests are all using hard coded poolState )

If no underlying token, this change just adds  `"underlyingToken": null` to `poolState.tokens[i]`

This approach would enable the same helper to be used for boosted and non boosted

### DIFF
--- a/src/data/providers/balancer-api/modules/pool-state/index.ts
+++ b/src/data/providers/balancer-api/modules/pool-state/index.ts
@@ -1,5 +1,8 @@
 import { BalancerApiClient } from '../../client';
-import { PoolState, PoolStateWithBalances } from '../../../../../entities';
+import {
+    PoolStateWithBalances,
+    PoolStateWithUnderlyings,
+} from '../../../../../entities';
 import { mapPoolType } from '../../../../../utils/poolTypeMapper';
 
 import { API_CHAIN_NAMES } from '../../../../../utils/constants';
@@ -16,6 +19,10 @@ export class Pools {
           index
           address
           decimals
+          underlyingToken {
+          address
+          decimals
+          }
         }
       }
     }`;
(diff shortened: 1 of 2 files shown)


brunoguerios:
I added a slightly different version of this on #584 - it creates a separate poolStateWithUnderlyings query. 
Since they are used in different scenarios, I think it's better DX than mixing both.
What do you think?

MattPereira:
@brunoguerios I agree your approach is better DX for us internally, but I kinda think it's worse DX for SDK consumer cus it forces dev to know and write logic to use a different query for boosted add vs standard add

Would be cool if we could figure out how to completely abstract away standard vs boosted add liquidity so devs don't have to worry about using `new AddLiquidityBoostedV3` and `Permit2Helper.signAddLiquidityBoostedApproval` for boosted pools

Franz just added flags that could make such a feat a bit easier
https://github.com/balancer/backend/blob/v3-canary/apps%2Fapi%2Fgql%2Fschema%2Fpool.gql#L1327-L1331

wdyt?

brunoguerios:
Interesting approach - so your suggestion is that we abstract away standard vs boosted add/remove liquidity operations depending on the user inputs being wrapped or underlying token addresses?
I like the overall idea, but it would be quite a refactor in the SDK - and a breaking change since we're changing external interface
Feels like something we could/should spend time once we cover more high priority tasks?
wdyt @johngrantuk ?

brunoguerios:
Ps: sorry, but I ended up merging the other solution alongside a hotfix FE was expecting 😬 
Since it's an api helper, I'd say it's ok to change it again once we reach a conclusion on how to approach the whole thing.

MattPereira:
@brunoguerios no sweat on the merge as the agent guy i'm supporting needed the sonic boosted router :raised_hands:

Issue #596, opened 2025-02-15
TokenAmount.toSignificant method fails to correctly encode high precision numbers


rafal2228:
Related code https://github.com/balancer/b-sdk/blob/ba2d86b4c42ea787ffc0419f2ff05b6a9766b330/src/entities/tokenAmount.ts#L83-L88

When integrating swap functionality with `fetchSorSwapPaths` function, I came across an issue with internal implementation using `toSignificant` method. After investigation turns out that referenced function fails to correctly encode high precision numbers.

Consider following unit test:

```ts
import { Token, TokenAmount } from '@balancer/sdk';
import { parseUnits } from 'viem';
import { describe, it, expect } from 'vitest';

const weth = new Token(
  42161,
  '0x82aF49447D8a07e3bd95BD0d56f35241523fBab1',
  18,
  'WETH',
);

describe('TokenAmount.toSignificant', () => {
  it('should correctly format small amounts of high precision tokens', () => {
    const amount = '0.000000312457062375';
    const parsed = parseUnits(amount, weth.decimals);

    const tokenAmount = TokenAmount.fromRawAmount(weth, parsed);

    expect(tokenAmount.toSignificant(weth.decimals)).toEqual(amount);
  });
});
```

Expected return value:
```ts
'0.000000312457062375' // same as initial input
```

Received:
```ts
'3.12457062375e-7'
```

This causes the underlying API used by `fetchSorSwapPaths` to return following error:
```json
{
    "data": null,
    "errors": [
        {
            "message": "Number `3.12457062375e-7` is not a valid decimal number.\n\nVersion: viem@2.22.16",
            "locations": [
                {
                    "line": 2,
                    "column": 3
                }
            ],
            "path": [
                "sorGetSwapPaths"
            ],
            "extensions": {
                "code": "INTERNAL_SERVER_ERROR"
            }
        }
    ]
}
```

PR #599, opened 2025-02-17
fix: TokenAmount.toSignificant with very small amounts

agualis:
Closes https://github.com/balancer/b-sdk/issues/596

Using   .toDecimalPlaces(significantDigits).**toString**();

was returning exponential notation when using very small numbers. Replacing `toString` by `toFixed` fixes the issue

### DIFF
--- a/.changeset/plenty-peaches-occur.md
+++ b/.changeset/plenty-peaches-occur.md
@@ -1,0 +1,5 @@
+---
+"@balancer/sdk": patch
+---
+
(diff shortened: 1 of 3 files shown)


brunoguerios (approved):
Awesome! Thanks for the fix 🙏

agualis:
@brunoguerios there was a flaky integration test. I re-run once and they passed (not related with this PR, I believe)

brunoguerios:
Yeah, we're aware of it and we're working to get it fixed. Thanks!

PR #602, opened 2025-02-24
Add monad testnet

### LINKED ISSUE #603: Update constants keys to be `ChainId` instead of `number`
That change would help us keep a better track of which chains we're supporting in the SDK and also to make sure we're not missing any configs/addresses for newly added chains.

MattPereira:
Closes #603 

Support for our friends on monad

### Summary
- Added `monadTestnet` as custom chain instead of bumping the viem version in case of breaking changes
- Used the same composite router address for boosted and nested version
- https://github.com/balancer/balancer-deployments/blob/monad-testnet/addresses/monad_testnet.json


src/utils/constants.ts
```diff
@@ -124,6 +127,7 @@ export const PERMIT2: Record<number, Address> = {
     [ChainId.SONIC]: '0x000000000022D473030F116dDEE9F6B43aC78BA3',
     [ChainId.ARBITRUM_ONE]: '0x000000000022D473030F116dDEE9F6B43aC78BA3',
     [ChainId.BASE]: '0x000000000022D473030F116dDEE9F6B43aC78BA3',
+    [ChainId.MONAD_TESTNET]: '0x000000000022D473030F116dDEE9F6B43aC78BA3',
```
MattPereira, line 130: I don't think this address is correct for monad testnet

https://testnet.monadexplorer.com/address/0x000000000022D473030F116dDEE9F6B43aC78BA3?tab=Activity

src/utils/customChains.ts
```diff
@@ -0,0 +1,29 @@
+import { defineChain } from 'viem';
+
+export const monadTestnet = /*#__PURE__*/ defineChain({
+    id: 10_143,
```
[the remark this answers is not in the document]
brunoguerios: Underscores in Typescript are just cosmetics to make it easier to read big numbers. They don't change number the value. Reference: https://mariusschulz.com/blog/numeric-separators-in-typescript

src/utils/constantsV3.ts
```diff
@@ -81,6 +87,7 @@ export const BALANCER_COMPOSITE_LIQUIDITY_ROUTER_NESTED: Record<
     [ChainId.SONIC]: '0xcf21664262774eBB2C2b559e13b47F6cA10F3E65',
     [ChainId.ARBITRUM_ONE]: '0x1311Fbc9F60359639174c1e7cC2032DbDb5Cc4d1',
     [ChainId.BASE]: '0xf23b4DB826DbA14c0e857029dfF076b1c0264843',
+    [ChainId.MONAD_TESTNET]: '0x36caC20dd805d128c1a6Dd16eeA845C574b5A17C',
```
MattPereira, line 90: I set this as same address as boosted, which is not the case on other networks where we are using two different versions of the composite router: one for nested and one for boosted

brunoguerios (approved):
LGTM ✅

brunoguerios:
Not necessarily related to this issue, but I created this one as an improvement suggestion: https://github.com/balancer/b-sdk/issues/603
Basically update constants to be `Record<ChainId, ...>` so we make sure we're not missing any config/address adding new chains.

PR #570, opened 2025-01-22
feat: create gyro E-CLP on v3

johngrantuk:
Closes #571 

### Summary
- refactored v3 pool creation inputs to all use `TokenConfig[]` type for the `tokens` property
- added `balancer-maths` as dependency for gyro ECLP input validations
  - https://github.com/balancer/balancer-maths/pull/97

PR #605, opened 2025-02-26
feat: create lbp

mkflow27:
This feature add lbp creation support to the sdk.


src/entities/createPool/types.ts
```diff
@@ -78,12 +78,42 @@ export type CreatePoolV3WeightedInput = CreatePoolV3BaseInput & {
...
+};
+
 export type CreatePoolInput =
     | CreatePoolV2WeightedInput
     | CreatePoolV2ComposableStableInput
     | CreatePoolV3WeightedInput
     | CreatePoolV3StableInput
-    | CreatePoolStableSurgeInput;
+    | CreatePoolStableSurgeInput
+    | CreateLiquidityBoostrappingPoolInput;
```
brunoguerios, line 116: I'd keep the standard:
```suggestion
    | CreatePoolLiquidityBootstrappingInput;
```

src/entities/inputValidator/composableStable/inputValidatorComposableStable.ts
```diff
@@ -34,7 +37,9 @@ export class InputValidatorComposableStable extends InputValidatorBase {
     }
 
     validateCreatePool(input: CreatePoolV2ComposableStableInput): void {
-        super.validateCreatePool(input);
+        validateCreatePoolTokens(input.tokens);
+        validateCreatePoolTokenConfig(input);
```
brunoguerios, line 41: `validateCreatePoolTokenConfig` seems to be relevant to v3 only and ComposableStable exists only on V2 so you should be able to remove this from here

src/entities/inputValidator/boosted/inputValidatorBoosted.ts
```diff
@@ -56,4 +64,15 @@ export class InputValidatorBoosted extends InputValidatorBase {
             }
         }
     }
+    validateCreatePool(
+        input:
+            | CreatePoolV2WeightedInput
+            | CreatePoolV2ComposableStableInput
```
brunoguerios, line 70: We don't have Boosted pools on V2, so you should be able to remove these 2 options

src/entities/inputValidator/liquidityBootstrapping/inputValidatorLiquidityBootstrapping.ts
```diff
@@ -0,0 +1,55 @@
...
+        if (input.lbpParams.startTime >= input.lbpParams.endTime) {
+            throw new Error('Start time must be before end time');
+        }
+        // cannot be in the past - technically allowed on the sc side.
+        // will simply move startTime to now.
+        if (input.lbpParams.startTime < BigInt(Math.floor(Date.now() / 1000))) {
+            throw new Error('Start time must be in the future');
+        }
+        // tokens cannot be the same
+        if (input.lbpParams.projectToken === input.lbpParams.reserveToken) {
```
brunoguerios, line 51: Please use `isSameAddress` helper to avoid issues with case sensitive comparisons

src/entities/inputValidator/utils/validateCreatePoolTokenConfig.ts
```diff
@@ -0,0 +1,32 @@
...
+    CreatePoolStableSurgeInput,
+} from '@/entities/createPool';
+
+import { TokenType } from '@/types';
+import { zeroAddress } from 'viem';
+
+export const validateCreatePoolTokenConfig = (
+    input:
+        | CreatePoolV2WeightedInput
+        | CreatePoolV2ComposableStableInput
```
brunoguerios, line 15: Same here about removing first 2 options (which are related to V2)

brunoguerios (commented):
Overall approach to input validator seems to make sense - I don't think we can do much better than this with how it's currently structured - let's consider a greater refactor in another PR  👍 
I left a few comments that should help polish the current implementation a bit.

MattPereira:
I'm having the same `Lint` check issues with #570 

maybe something to do with that last PR bruno merged? :thinking:

pnpm-lock.yaml
johngrantuk, line 1: Should be no change to the pnpm-lock file.

johngrantuk:
> I'm having the same `Lint` check issues with #570
> 
> maybe something to do with that last PR bruno merged? 🤔

I believe its because the pnpm-lockfile is updated with a different version. Reverting that change should fix things.

pnpm-lock.yaml
MattPereira: @mkflow27 I ran into the same `lockfileVerison: '9.0'` change causing problem with `Lint` check

The cause was that my machine had pnpm `v9.x `globally installed

The fix was this

1. global install of verison that matches our `package.json`

```
npm install -g pnpm@8.6.0
```

2. revert the lock file back to state of latest commit on main branch 
```
git checkout de8956de -- pnpm-lock.yaml
```

3. run install again
```
pnpm install
```

johngrantuk (changes requested):
I believe #644 should be merge/released soon so will probably require some updates after that (and a good test for the script too to see how handle the LBP additions).

src/abi/liquidityBootstrappingFactory.ts
MattPereira, line 1: this ABI already exists at `/abi/v3/lBPoolFactory.ts` exported as `lBPoolFactoryAbi_V3`

i do prefer how you named the abi "liquidityBoostrapping" since `lBP` is gross to read ( is generated by script to be consistent camel case with all the other abi export names )

MattPereira:
Do we know if LBP will be deployed to sonic?

MattPereira (approved):
Nice job on this one! :clap: 

I'm fine with however you decide to handle the lbp factory abi

src/entities/createPool/types.ts
```diff
@@ -146,3 +147,33 @@ export type CreatePoolReClammInput = Omit<
...
+    projectTokenStartWeight: bigint;
+    reserveTokenStartWeight: bigint;
+    projectTokenEndWeight: bigint;
+    reserveTokenEndWeight: bigint;
+    startTime: bigint;
+    endTime: bigint;
+    blockProjectTokenSwapsIn: boolean;
+};
+
+// The pool uses default liquidity management values (no setters available)
```
johngrantuk, line 164: Really useful comments 👏

src/entities/inputValidator/boosted/inputValidatorBoosted.ts
```diff
@@ -65,4 +71,13 @@ export class InputValidatorBoosted extends InputValidatorBase {
             }
         }
     }
+    validateCreatePool(
+        input:
+            | CreatePoolV3WeightedInput
```
johngrantuk, line 76: Why are these specific pool types included but e.g. gyro is not?

src/entities/createPool/createPoolV3/liquidityBootstrapping/createLiquidityBootstrapping.ts
```diff
@@ -0,0 +1,40 @@
+import { getRandomBytes32 } from '@/entities/utils/getRandomBytes32';
+import { encodeFunctionData } from 'viem';
+
+import {
+    CreatePoolBase,
+    CreatePoolBuildCallOutput,
+    CreatePoolLiquidityBootstrappingInput,
+} from '../../types';
+
+import { lBPoolFactoryAbi_V3 } from '@/abi';
```
johngrantuk, line 10: Can you change to use "extended" abi - check out the other implementations as an example.

test/lib/utils/createPoolHelper.ts
```diff
@@ -32,6 +33,7 @@ export async function doCreatePool(
             [PoolType.StableSurge]: stableSurgeFactoryAbiExtended,
             [PoolType.GyroE]: gyroECLPPoolFactoryAbiExtended,
             [PoolType.ReClamm]: reClammPoolFactoryAbiExtended,
+            [PoolType.LiquidityBootstrapping]: lBPoolFactoryAbi_V3,
```
johngrantuk, line 36: can change to extended ABI to match others.

src/entities/inputValidator/liquidityBootstrapping/inputValidatorLiquidityBootstrapping.ts
```diff
@@ -0,0 +1,82 @@
...
+import {
+    RemoveLiquidityInput,
+    RemoveLiquidityRecoveryInput,
+} from '../../removeLiquidity/types';
+import { CreatePoolLiquidityBootstrappingInput } from '../../createPool/types';
+
+import { isSameAddress, SDKError } from '@/utils';
+
+export class InputValidatorLiquidityBootstrapping extends InputValidatorBase {
+    // TODO
```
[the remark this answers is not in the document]
mkflow27: In the past we agreed that this pr should only be for creating lbps. Any further stuff should be added in a seperate pr. I'll create issues for it.

test/lib/utils/createPoolHelper.ts
mkflow27: addressed

src/entities/createPool/createPoolV3/liquidityBootstrapping/createLiquidityBootstrapping.ts
mkflow27: addressed

src/entities/inputValidator/boosted/inputValidatorBoosted.ts
mkflow27: So Gyro is excluded it because this InputValidator will not accept poolTypes, which are not Boosted. However All of the three types below can be boosted pools. 

I will need to do more investigation here for gyro pools.

This validator will get used if the PoolType.Boosted is selected. This also means that I need to actually additionally check if functionality which is used in the "InputValidatorWeighted" is properly reused in the InputValidatorBoosted. 

same thing for Gyro in that case. Say a GyroPool would be considered boosted, I would need to do the GyroValidation as well as the boosted validation (in case there is any) for the pool.

mkflow27: Adding some more thoughts on this change. If I understood it correctly: before this change a PoolType.Boosted would'vegone through the InputValidatorBase `validateCreatePool`. Sice this pr moves creation validation from the InputValidatorBase to each invidivual Inputvalidator. I think the reasoning here as also based off conversation with Bruno that we would only consider Weighted, Stable, Stablesurge as potentially boosted pools.

So I think the decision that needs to be made for this pr is either:
leave it as `CreatePoolInput` from the InputValidatorBase or leave it with these three types. 

If I leave it with `CreatePoolInput` there would be less changes in this pr. Let me know what you think.

But also based off my previous comment, the InputValidators need updating I believe but I feel this would be a separate issue.

src/entities/inputValidator/gyro/inputValidatorGyro.ts
```diff
@@ -6,15 +6,19 @@ import {
...
+import { validateCreatePoolTokenConfig } from '../utils/validateCreatePoolTokenConfig';
+
 import { inputValidationError, poolTypeError } from '@/utils';
 import { CreatePoolGyroECLPInput } from '@/entities/createPool';
 import { GyroECLPMath } from '@balancer-labs/balancer-maths';
 export class InputValidatorGyro extends InputValidatorBase {
     validateCreatePool(input: CreatePoolGyroECLPInput) {
-        super.validateCreatePool(input);
+        validateCreatePoolTokenConfig(input);
+        validateCreatePoolTokens(input.tokens);
```
brunoguerios, line 21: If I understood correctly, you moved these to each pool type specific implementations because you won't use this within the LBP implementation, right?
You could leave them in the `InputValidatorBase` and simply not call `super.validateCreatePool(input);` within the LBP implementation, which should achieve the same end result with less code duplication.
Should be ok to leave as is as well. Not a big deal.

test/v3/addLiquidity/addLiquidity.liquiditybootstrapping.integration.test.ts
```diff
@@ -0,0 +1,281 @@
...
+    let lbpParams: LBPParams;
+    let createPoolInput: CreatePoolLiquidityBootstrappingInput;
+    let poolAddress: Address;
+    let testAddress: Address;
+
+    beforeAll(async () => {
+        ({ rpcUrl } = await startFork(
+            ANVIL_NETWORKS[ChainId[chainId]],
+            undefined,
+            7831624n,
```
brunoguerios, line 65: Do you really need a separate blockNumber for this test? 
Ideally you should be able to update the global blockNumber so all tests run in the same block and we benefit from fork cache speeding up tests.

test/v3/removeLiquidity/removeLiquidity.liquiditybootstrapping.integration.test.ts
```diff
@@ -0,0 +1,406 @@
...
+    let poolAddress: Address;
+    let testAddress: Address;
+    let startTime: bigint;
+    let endTime: bigint;
+
+    beforeAll(async () => {
+        ({ rpcUrl } = await startFork(
+            ANVIL_NETWORKS[ChainId[chainId]],
+            undefined,
+            7831624n,
```
brunoguerios, line 71: Same thing here about the blockNumber.

brunoguerios (commented):
I left a few comments I'd like to discuss before approving/merging, but overall looks good. 👍

src/entities/inputValidator/gyro/inputValidatorGyro.ts
mkflow27: The main reason I moved them is because The `InputValidatorBase` was accessing the `tokens` of the `reatePoolInput` type. `tokens` however do not exist on the `CreatePoolLiquidityBootstrappingInput` type creating a mismatch. Therefor I had to move the create pool validation to each individual pool type validator.

test/v3/addLiquidity/addLiquidity.liquiditybootstrapping.integration.test.ts
mkflow27: I don't need them. I initially planned to add liq to an already deployed lbp. The test now however creates a lbp instead of using an existing one.

test/v3/removeLiquidity/removeLiquidity.liquiditybootstrapping.integration.test.ts
mkflow27: I don't need them. I initially planned to add liq to an already deployed lbp. The test now however creates a lbp instead of using an existing one.

test/v3/addLiquidity/addLiquidity.liquiditybootstrapping.integration.test.ts
```diff
@@ -0,0 +1,253 @@
...
+            testAddress,
+            addLiquidityBuildCallOutput.to,
+            addLiquidityBuildCallOutput.callData,
+            addLiquidityBuildCallOutput.value,
+        );
+
+        // tokens leaver wallet, bpt enter wallet (all deltas are positive)
+        txOutput.balanceDeltas.forEach((balance) => {
+            expect(balance).toBeGreaterThan(0n);
+        });
```
brunoguerios, line 251: I'm leaning towards not having these tests at all, but if we do keep them, you should compare query with actual transaction results. Simply checking if all is positive isn't enough.
There are helpers to assert add liquidity operations you could reuse.

test/v3/removeLiquidity/removeLiquidity.liquiditybootstrapping.integration.test.ts
```diff
@@ -0,0 +1,345 @@
...
+            removeLiquidityBuildCallOutput.to,
+            removeLiquidityBuildCallOutput.callData,
+            removeLiquidityBuildCallOutput.value,
+        );
+        expect(removeLiqTxOutput.transactionReceipt.status).toBe('success');
+        expect(
+            removeLiqTxOutput.balanceDeltas.forEach((balance) => {
+                expect(balance).toBeGreaterThan(0n);
+            }),
+        );
```
brunoguerios, line 343: Same here.

test/v3/addLiquidity/addLiquidity.liquiditybootstrapping.integration.test.ts
mkflow27: I have removed the tests

test/v3/removeLiquidity/removeLiquidity.liquiditybootstrapping.integration.test.ts
mkflow27: I have removed the tests

src/entities/inputValidator/boosted/inputValidatorBoosted.ts
johngrantuk: Tracking in: https://github.com/balancer/b-sdk/issues/673

johngrantuk (approved):
LGTM

src/entities/inputValidator/liquidityBootstrapping/inputValidatorLiquidityBootstrapping.ts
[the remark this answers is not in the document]
mkflow27: Tracking it here: https://github.com/balancer/b-sdk/issues/674

PR #606, opened 2025-02-26
Fix test failing intermittently

### LINKED ISSUE #565: Fix add liquidity proportional integration test failing on CI
A single integration test scenario running on gnosis chain was added.
It passes locally, but fails on CI, apparently due to some missing config (more details [here](https://github.com/balancer/b-sdk/actions/runs/12835861369/job/35796261816?pr=556))
Currently the test is being skipped so the fix is pushed to prod, but it would be nice to investigate and get it working on CI as well

Ps: adding liquidity proportionally to any boosted pool with their wrapped tokens (instead of underlying ones) should be enough to test this implementation - in other words, using regular router instead of the composite router

brunoguerios:
- Fix Add Liquidity Boosted Integration test that was intermittently failing on CI

Also:
Closes #565

### DIFF
--- a/test/v3/addLiquidity/addLiquidity.integration-gnosis.test.ts
+++ b/test/v3/addLiquidity/addLiquidity.integration-gnosis.test.ts
@@ -51,7 +51,7 @@ const sDAI = TOKENS[chainId].sDAI;
 const BRLA = TOKENS[chainId].BRLA;
 
 // skipping until we figure out why test is failing on CI
-describe.skip('add liquidity test', () => {
+describe('add liquidity test', () => {
     let client: PublicWalletClient & TestActions;
     let txInput: AddLiquidityTxInput;
     let poolState: PoolState;
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
         "@types/async-retry": "^1.4.8",
         "@types/lodash.clonedeep": "^4.5.9",
         "@types/node": "^18.19.7",
-        "@viem/anvil": "^0.0.6",
+        "@viem/anvil": "^0.0.10",
         "dotenv": "^16.3.1",
         "dpdm": "^3.14.0",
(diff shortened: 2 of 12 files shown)


brunoguerios:
Main change: adapt anvil fork to mine blocks every 1 second, instead of default behavior of instant auto-mine after every transaction.

Worth noting that this caused issues with how we read balances to check for success. Apparently `getBalance` reads are applied to the last block mined, so we need to wait for confirmation of the token balance setup (i.e. stataUSDT minting on this specific test) in order to properly read balances and assert results.

Also, apparently `snapshot` doesn't work well with these time based blocks, so these tests are "incremental" in the sense that they run one after the other in the same local fork.

MattPereira:
I think this is just problem with my machine that I will investigate, but wanted to report just in case

For some reason takes more than twice as long for me to run locally vs the GHA

image

mkflow27 (approved):
Great changes Bruno.

test/v3/addLiquidity/addLiquidity.integration-gnosis.test.ts
```diff
@@ -51,7 +51,7 @@ const sDAI = TOKENS[chainId].sDAI;
 const BRLA = TOKENS[chainId].BRLA;
 
 // skipping until we figure out why test is failing on CI
```
johngrantuk, line 53: Remove this comment

brunoguerios:
I'm tracking down the source of the issue here. Apparently the issue lies in the `fallBackRpc` url, because the one I have set on .env is working as expected. I'll fix that over here.

brunoguerios:
Sharing what I found so you're aware:

>attempting to fetch an env var as `process.env[<<key>>]` for a non existing key will sometimes return `undefined` and sometimes `'undefined'` as string - we have to check both cases before deciding to use the `fallBackRpc`

PR #616, opened 2025-03-12
Calculate derived params helper for E-CLP

MattPereira:
### Summary
- scale base params from 18 to 100, do some maths, return derived params in scale 38
- used Big.js to do the square root because [GyroPoolMath.sqrt](https://github.com/balancer/balancer-maths/blob/876bdc75c884f52d302d850dcf6721cd6cc7b898/typescript/src/gyro/gyroPoolMath.ts#L15-L49) doesn't seem compatible with precision 100
- Used params from GyroECLP pools [listed on beets.fi](https://beets.fi/pools?poolTypes=GYRO) for extra test cases


brunoguerios (commented):
LGTM ✅ 
Regarding adding a dependency, I'd say it's ok.
My concerns would be:
- safety: big.js seems like a well known package with ~22M weekly downloads, so I wouldn't expect any issues here
- size (so it doesn't bloat SDK package): it's quite small ~50kb, so no issues here as well

mkflow27 (approved):
LGTM

test/v3/createPool/gyroECLP/gyroECLP.integration.test.ts
```diff
@@ -132,11 +133,73 @@ describe('GyroECLP - create & init', () => {
         });
     }, 120_000);
 
-    test('pool should be created', async () => {
+    test('calc derived params #1', async () => {
```
johngrantuk, line 136: the derived tests should probably live in a separate test outwith the integration test?

src/entities/createPool/createPoolV3/gyroECLP/createPoolGyroECLP.ts
```diff
@@ -85,3 +86,75 @@ export class CreatePoolGyroECLP implements CreatePoolBase {
...
+        u: (u * D38) / D100,
+        v: (v * D38) / D100,
+        w: (w * D38) / D100,
+        z: (z * D38) / D100,
+        dSq: (dSq * D38) / D100,
+    };
+}
+
+function bigIntSqrt(val: bigint): bigint {
+    return BigInt(new Big(val.toString()).sqrt().toFixed(0));
```
johngrantuk, line 159: Would the existing [gyro math sqrt implementation](https://github.com/balancer/balancer-maths/blob/main/typescript/src/gyro/gyroPoolMath.ts#L15) work?

test/v3/createPool/gyroECLP/gyroECLP.integration.test.ts
MattPereira: ya that makes sense ✅

src/entities/createPool/createPoolV3/gyroECLP/createPoolGyroECLP.ts
MattPereira: I tried to use that sir isaac newton style sqrt first, but it runs into the `throw Error('_sqrt FAILED')` when feeding a 100 decimal number as the input with any sane amount of tolerance. ( it does succeed with tolerance `1000000000000000000000000000000000000000000000000000000000000n` :joy: )

But our AI friends think that it is theoretically possible to modify that `GyroPoolMath.sqrt` function to work with 1e100 scale :thinking:

johngrantuk: Definitely don't want to modify `GyroPoolMath.sqrt` as it needs to match the smart contract implementation.
Looks like we will live with the dependencie addition.

johngrantuk (approved):

brunoguerios (approved):

PR #626, opened 2025-03-20
Remove test no longer necessary on gnosis-chain

brunoguerios:
This test was added to make sure a scenario with live balances was being covered, but now it's no longer needed and it's causing issues with CI due to RPC on gnosis being unstable. Since it's the only test on gnosis, I believe it's more beneficial to remove it than trying to find a better RPC.


MattPereira (approved):
I concur ✅

PR #628, opened 2025-03-21
Refactor price impact errors as SDKError

### LINKED ISSUE #619: Improve Error Handling - Price Impact Errors
2. for “complex” operations, which involve multiple underlying on-chain queries (e.g. price impact calculation), SDK should provide custom/standardized error messages with proper context
    1. Suggested standard:
        1. SDK Error - [operation] - [error message]
    2. Example:
        1. SDK Error - Price Impact - Add liquidity unbalanced boosted failed with unexpected error while calculating token deltas

More details here: https://www.notion.so/SDK-Error-Handling-1b4e395e228180ca8eeaf50ac3584ccb?pvs=4

brunoguerios:
Close #619

### DIFF
--- a/src/entities/priceImpact/addLiquidityUnbalanced.ts
+++ b/src/entities/priceImpact/addLiquidityUnbalanced.ts
@@ -1,4 +1,6 @@
-import { abs, max, min } from '@/utils';
+import { abs, max, min, SDKError } from '@/utils';
+import { SwapKind } from '@/types';
+
 import { AddLiquidity } from '../addLiquidity';
 import { AddLiquidityUnbalancedInput } from '../addLiquidity/types';
 import { PriceImpactAmount } from '../priceImpactAmount';
@@ -11,7 +13,6 @@ import { Token } from '../token';
 import { TokenAmount } from '../tokenAmount';
 import { PoolState } from '../types';
 import { priceImpactABA } from './helper';
-import { SwapKind } from '@/types';
 import { Swap, SwapInput } from '../swap';
 
 export const addLiquidityUnbalanced = async (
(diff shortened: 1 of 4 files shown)


src/entities/priceImpact/addLiquidityUnbalanced.ts
```diff
@@ -171,14 +170,22 @@ export const addLiquidityUnbalanced = async (
...
+                },
+                poolState,
+            );
+            const signal = deltas[tokenIndex] >= 0n ? 1n : -1n;
+            return deltaBPT.amount * signal;
+        } catch {
+            throw new SDKError(
+                'Price Impact',
+                'Add Liquidity Unbalanced',
+                'Unexpected error while calculating addLiquidityUnbalanced PI at Delta add step',
```
MattPereira, line 187: Would it be helpful for debugging if we `catch (err)` and pass it through here ?

MattPereira (approved):
LGTM ✅

mkflow27 (approved):
LGTM

src/entities/priceImpact/addLiquidityUnbalanced.ts
brunoguerios: That was intentionally removed.
1. The SC error in this case is not so relevant to the FE (and might be misleading)
2. Removing this currently allows FE to differentiate PriceImpact errors from "pure" SC error more easily, because they rely on parsing error `message` for that - after this refactor, I'll suggest they use error `name` instead.
3. Locally for debugging it might be helpful, but I'd say it's better for us to include when we need it instead of the other way around.

MattPereira: ah that makes sense thanks for explaining the context! :pray:

PR #634, opened 2025-03-30
Avalanche Support

### LINKED ISSUE #633: Avalanche Support

mkflow27:
Fixes #633

This pr adds avalanche support with nested operations blocked. Also stable surge pool creation is blocked as the hook & new factory is not yet deployed.

The check for chainId happens as part of `AddLiquidityNestedV3` and `RemoveLiquidityNesteV3` each time as part of the `query` and `buildCall operations`

Checks for signatures are excluded as eventually the `query` and `buildCall` operations are used and the check trigger the throwing of an error.


MattPereira:
@mkflow27 apologies but i can't recall the details since the slack message i linked to auto-deleted :upside_down_face: 
- https://github.com/balancer/b-sdk/pull/572

Maybe is fine to just use the sole composite deployed to Avalanche for both?

src/entities/addLiquidityNested/index.ts
```diff
@@ -33,6 +33,9 @@ export class AddLiquidityNested {
                 return addLiquidity.query(input, nestedPoolState);
             }
             case 3: {
+                if (input.chainId === 43114) {
```
MattPereira, line 36: Maybe consider using `ChainId.AVALANCHE` instead of number for improved readability?

src/entities/addLiquidityNested/index.ts
```diff
@@ -33,6 +33,9 @@ export class AddLiquidityNested {
...
+                    throw new Error('Avax not supported for nested operations');
```
MattPereira, line 37: Could be confusing since `AVAX` is also a ticker :sweat_smile: 

Maybe use the new SDK error like..

```
                if (input.chainId === ChainId.AVALANCHE) {
                    throw new SDKError(
                        'Input Validation',
                        'Add Liquidity Nested',
                        'Balancer V3 does not support this operation on Avalanche',
                    );
                }
```

src/entities/addLiquidityNested/index.ts
mkflow27: Good suggestion, added

src/entities/addLiquidityNested/index.ts
```diff
@@ -33,6 +34,13 @@ export class AddLiquidityNested {
                 return addLiquidity.query(input, nestedPoolState);
             }
             case 3: {
+                if (input.chainId === ChainId.AVALANCHE) {
+                    throw new SDKError(
+                        'Input Validation',
+                        'Add Liquidity Nested',
+                        'Balancer V3 does not support this operation on Avalanche',
```
brunoguerios, line 41: Mind that consumers will usually show just the error message, so having the actual action within it should be helpful:
```suggestion
                        'Balancer V3 does not support Add Liquidity Nested on Avalanche',
```

src/entities/removeLiquidityNested/index.ts
```diff
@@ -48,6 +55,13 @@ export class RemoveLiquidityNested {
                 return removeLiquidity.buildCall(input);
             }
             case 3: {
+                if (input.chainId === ChainId.AVALANCHE) {
+                    throw new SDKError(
+                        'Input Validation',
+                        'Add Liquidity Nested',
+                        'Balancer V3 does not support this operation on Avalanche',
```
brunoguerios, line 62:
```suggestion
                        'Remove Liquidity Nested',
                        'Balancer V3 does not support Remove Liquidity Nested on Avalanche',
```

brunoguerios (approved):

PR #636, opened 2025-04-01
Fix map pool type

johngrantuk:
V3 does not need a specific pool type for add/removes/swaps to work. We were using `mapPoolType` in API provider which would throw if we don't manually add each new type to supported types. I've changed it to just use the type passed (as string) whenever its V3.


mkflow27 (approved):
Mapping of pool type only on V2 now. LGTM.

brunoguerios (approved):
LGTM ✅

PR #642, opened 2025-04-02
Cache node_modules before install

### LINKED ISSUE #639: Investigate why Install check started failing
We want to know why GHA cloud stopped caching our packages and potentially explore a better solution

See #638 for context. 

Maybe it is better to always install before caching for reliability / consistency reasons?

MattPereira:
Closes #639 

Per @johngrantuk 
> Looks like npm had an outage yesterday which might be the reason: https://status.npmjs.org/

### Summary
- Moved install after cache lookup for node_modules

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -16,12 +16,13 @@ runs:
         path: ~/.pnpm-store
         key: ${{ matrix.node-version }}-pnpm-${{ hashFiles('pnpm-lock.yaml') }}
         restore-keys: ${{ matrix.node-version }}-pnpm-
-    - name: Install Dependencies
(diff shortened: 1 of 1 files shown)


brunoguerios (approved):
👍

brunoguerios:
Do we need a changeset for this? 🤔 
I'm assuming we don't, but thought it would make sense to raise the question.

johngrantuk:
> Do we need a changeset for this? 🤔 I'm assuming we don't, but thought it would make sense to raise the question.

Think changesets are only needed for the prod code rather than CI.

johngrantuk (approved):
Lets give it a shot.

PR #644, opened 2025-04-03
Script for updating deployments

### LINKED ISSUE #643: Automate New Network
* Use a local SDK script to pull addresses from deployments repo and generate PR from template
* PR would then be opened manually (but open to auto if possible)

MattPereira:
Closes #643 

### Summary
- Fetch all `ACTIVE` v2 and v3 contracts from [balancer-deployments/addresses](https://github.com/balancer/balancer-deployments/tree/master/addresses) 
  - Save addresses for each network to `balancerV2Contracts.ts` and `balancerV3Contracts.ts`
  - Save abis to `src/abi/v2` and `src/abi/v3/`
- Fetch sonic addresses from beets org at [balancer-deployments/addresses/sonic.json](https://github.com/beethovenxfi/balancer-deployments/blob/master/addresses/sonic.json)
- Replaced all the v3 contract addresses and ABIs with data generated by script
- Created issue to update V2 addys later #650 
- Appending `_V2` and `_V3` to ABIs to avoid any naming conflicts and make it clear which protocol version

### Questions

- Should the SDK support addresses that are not ACTIVE in deployments repo?
  - Only one `CompositeLiquidityRouter` is ACTIVE in the deployments repo
  -  Leave `BALANCER_COMPOSITE_LIQUIDITY_ROUTER_NESTED` untouched for now?
- Where did cowAMMPool ABI come from?

### TODO
- [x] Pull in the sonic addresses from [beets deployments repo](https://github.com/beethovenxfi/balancer-deployments/blob/master/addresses/sonic.json)
- [x] ask daniel to confirm we can rely on the beets org addys on sonic to stay in sync with ABIs taken from balancer org


johngrantuk:
Looks like its easy to pull the addresses. Can we take it a step further and try to automate something like this [PR](https://github.com/balancer/b-sdk/pull/602)? I think avoiding GHA is the right approach for now.

src/utils/constantsV3.ts
```diff
@@ -1,11 +1,13 @@
...
 
+const { Vault } = balancerV3Contracts;
 // V3
 export const VAULT_V3: Record<number, Address> = {
-    [ChainId.SEPOLIA]: '0xbA1333333333a1BA1108E8412f11850A5C319bA9',
-    [ChainId.MAINNET]: '0xbA1333333333a1BA1108E8412f11850A5C319bA9',
-    [ChainId.GNOSIS_CHAIN]: '0xbA1333333333a1BA1108E8412f11850A5C319bA9',
+    [ChainId.SEPOLIA]: Vault[ChainId.SEPOLIA],
+    [ChainId.MAINNET]: Vault[ChainId.MAINNET],
+    [ChainId.GNOSIS_CHAIN]: Vault[ChainId.GNOSIS_CHAIN],
```
MattPereira, line 10: We could do this but I kinda like abandoning the `constantsV3.ts` file entirely since all ACTIVE v3 addresses will be in `balancerV3Contracts`

src/utils/tokens.ts
```diff
@@ -1,7 +1,7 @@
 import { Address, Client, Hex, PublicClient, getContract } from 'viem';
 import { vaultV2Abi, vaultExtensionAbi_V3 } from '../abi';
-import { VAULT_V2 } from './constantsV2';
-import { VAULT_V3 } from './constantsV3';
+import { balancerV2Contracts } from './balancerV2Contracts';
+import { balancerV3Contracts } from './balancerV3Contracts';
```
MattPereira, line 4: Example of what i mean above ^

src/abi/index.ts
```diff
@@ -10,8 +10,8 @@ import { gyroECLPPoolFactoryAbi_V3 } from './gyroECLPPoolFactory.V3';
 import { stablePoolFactoryAbi_V3 } from './stablePoolFactory.V3';
 import { stableSurgeFactoryAbi } from './stableSurgeFactory';
 import { permit2Abi } from './permit2';
-import { vaultV2Abi } from './vault.V2';
-import { vaultV3Abi } from './vault.V3';
+import { vaultAbi as vaultV2Abi } from './v2';
+import { vaultAbi as vaultV3Abi } from './v3';
```
MattPereira, line 14: Example of what replacing manually copied ABIs with script ABIs might look like

scripts/updateDeployments.ts
MattPereira, line 1: Script fetches all the addresses from the deployments-repo and uses taskIDs to grab ABIs out of the artifacts

src/utils/tokens.ts
mkflow27: I like the approach importing from `balancerV3Contracts` more than importing `VAULT_V3` which imports from `balancerV3Contracts`

scripts/updateDeployments.ts
```diff
@@ -0,0 +1,136 @@
...
+    const balancerV3Contracts: ContractRegistry = {};
+
+    await Promise.all(
+        supportedNetworks.map(async ({ name: networkName, chainId }) => {
+            const url = `https://raw.githubusercontent.com/balancer/balancer-deployments/refs/heads/master/addresses/${networkName}.json`;
+            const res = await fetch(url);
+            const data: NetworkRegistryResponse = await res.json();
+
+            Object.entries(data)
+                .filter(([_, value]) => value.status === 'ACTIVE')
```
brunoguerios, line 73: I feel like it's a bit weird to have all of those addresses not relevant to the SDK being pulled over here.
Can we also filter by the `targetContracts` arrays above? Then we'll make sure we'll only register addresses relevant to the SDK in the repo 🙃

src/utils/tokens.ts
brunoguerios: Same here 👍
Importing directly from balancerContracts files seem to make more sense.

brunoguerios (commented):
Awesome to see this kind of thing being automated 👏

src/entities/permitHelper/index.ts
```diff
@@ -176,7 +175,9 @@ const signPermit = async (
 const getDomain = async (client: PublicWalletClient, token: Hex) => {
     const [, name, version, chainId, verifyingContract, , ,] =
         await client.readContract({
-            abi: weightedPoolAbi_V3,
+            abi: parseAbi([
+                'function eip712Domain() view returns (bytes1 fields, string name, string version, uint256 chainId, address verifyingContract, bytes32 salt, uint256[] extensions)',
+            ]),
```
MattPereira, line 180: @brunoguerios

Think this may be better since this was only place `weightedPoolAbi_V3` was being used and it doesn't seem to be directly related to `getDomain` function or `PermitHelper` class

Happy to revert if team prefers tho. Assuming the abi came from a `MockStablePool` deployment?

johngrantuk (approved):
Lets give it a go!
As highlighted by Alberto/MkFlow in sdk-frontend channel the FE does [import addresses](https://github.com/balancer/frontend-monorepo/blob/680334b6c4d2444d3b82f17be0e7af5878acff80/packages/lib/config/networks/mainnet.ts#L7) from the SDK so can you also manage that before release?

src/entities/permitHelper/index.ts
brunoguerios: I used the pool abi initially for "simplicity", but I agree it's not pool specific and looks better with the explicit abi like this 👍

src/utils/balancerV2Contracts.ts
```diff
@@ -0,0 +1,92 @@
+export const balancerV2Contracts = {
+    Authorizer: {
+        '1': '0xA331D84eC860Bf466b4CdCcFb4aC09a1B43F3aE6',
+        '10': '0xA331D84eC860Bf466b4CdCcFb4aC09a1B43F3aE6',
+        '56': '0xA331D84eC860Bf466b4CdCcFb4aC09a1B43F3aE6',
+        '100': '0xA331D84eC860Bf466b4CdCcFb4aC09a1B43F3aE6',
```
brunoguerios, line 6: I usually find it a bit annoying to read through ids like these 😬 
Being able to read chain names during a sanity check on the addresses is a better experience.
If it's a small change in the script, I'd say it would be _nice to have_, but if it ends up being a pain or if it requires updating multiple files, I'd say it's ok as is.

scripts/updateDeployments.ts
```diff
@@ -0,0 +1,213 @@
...
+                                            contract.address;
+                                    }
+                                    if (version === 'v3') {
+                                        if (!balancerV3Contracts[name])
+                                            balancerV3Contracts[name] = {};
+                                        balancerV3Contracts[name][chainId] =
+                                            contract.address;
+                                    }
+
+                                    // grab contract abis using only mainnet to avoid redundant requests
```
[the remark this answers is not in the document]
MattPereira: ya i am concerned about this part of script potentially causing trouble some day

src/utils/balancerV2Contracts.ts
MattPereira: Was a bit tricky but definitely worth it :raised_hands: 

Make it alot easier to sanity check and see things like no `LBPPoolFactory` is deployed to Sonic yet

brunoguerios: Thanks for making the change! 🙏 
One last thing about them -> I see they are now sorted differently then before.
I'm just wondering if that order is random by any chance. If yes, it would mean that running the script to update addresses would generate a different order, which is not ideal for evaluating diffs.
I'm ok on sorting by id or alphabetically by chain name, but it would be great if that was consistent.

PR #646, opened 2025-04-04
Attempt to fix swap v2 integration test failing on CI

### LINKED ISSUE #645: SwapV2 test intermittently failing on CI
From @MattPereira 

> Had a test fail during the CI run that is triggered post merge
https://github.com/balancer/b-sdk/actions/runs/13776222277/job/38526017933

and

> same swap v2 post merge fail. i may be cursed :upside_down_face:
https://github.com/balancer/b-sdk/actions/runs/13912865419/job/38930490613

brunoguerios:
Closes #645 

Main changes:
The assumption is that sequential tests may have been conflicting with each other based on the fact that they run in parallel and reuse the same fork node.
In order to fix that it was applied the pattern from other working tests, where the fork setup happens only once and then each test restarts it from a snapshot.

Extra changes:
- In order to avoid the need to fix the blockNumber, tests comparing against hardcoded amounts were removed. Since they were checking against query results, they were considered redundant when compared to tests checking actual transaction results.

Ps: no changeset needed since this test fix does not require a package release

### DIFF
--- a/test/entities/swaps/v2/swapV2.integration.test.ts
+++ b/test/entities/swaps/v2/swapV2.integration.test.ts
@@ -9,41 +9,39 @@ import {
     publicActions,
     walletActions,
     TestActions,
+    Hex,
 } from 'viem';
 import {
     CHAINS,
     ChainId,
     SwapKind,
-    Token,
     Swap,
-    ExactInQueryOutput,
-    ExactOutQueryOutput,
     VAULT_V2,
     Path,
     PublicWalletClient,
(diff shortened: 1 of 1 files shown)


MattPereira (approved):
Agree with the assumption and attempt to fix 

Was maybe some sort of race condition?

Much appreciated! :pray:

mkflow27 (approved):
Thanks for making these changes

johngrantuk (approved):
Lets give it a try!

PR #658, opened 2025-04-14
feat: create ReClamm pools

### LINKED ISSUE #657: Add reclamm pool creation
Coordinate with Bruno on how to best handle the quirky init

### Resources
- https://github.com/balancer/balancer-deployments/pull/235
- https://github.com/balancer/reclamm
- https://balancerecosystem.slack.com/archives/C08DTJ8SFUY
- https://www.notion.so/reCLAMM-Requirements-19ee395e22818033a4fef9618afd6e9e

MattPereira:
Closes #657 

### Summary
- Refactored create pool v3 input types to be more consistent / maintainable
- Implemented `computeInitialBalanceRatio` in balancer maths repo so we dont have to use on chain fetch [see commit](https://github.com/balancer/balancer-maths/pull/101/commits/816aabddb3ad49bbe22fe1aca5884f7f0e060167)
- Utility helper function to `calculateReClammInitAmounts` using a given token amount in


src/entities/createPool/types.ts
```diff
@@ -105,15 +83,66 @@ export type CreatePoolV2ComposableStableArgs = [
...
+///// Balancer v3 /////
+export type CreatePoolV3BaseInput<T = TokenConfig> = CreatePoolBaseInput & {
+    protocolVersion: 3;
+    pauseManager: Address;
+    swapFeeManager: Address;
+    swapFeePercentage: bigint;
+    poolHooksContract: Address;
+    enableDonation: boolean;
+    disableUnbalancedLiquidity: boolean;
+    tokens: T[];
```
MattPereira, line 95: TIL about generic type parameters with a default type! :cool:

src/entities/createPool/types.ts
```diff
@@ -105,15 +83,66 @@ export type CreatePoolV2ComposableStableArgs = [
...
-export type PoolRoleAccounts = {
-    pauseManager: Address;
-    swapFeeManager: Address;
-    poolCreator: Address;
+export type TokenConfigWithWeight = TokenConfig & {
+    weight: bigint;
+};
+
+export type CreatePoolV3WeightedInput =
+    CreatePoolV3BaseInput<TokenConfigWithWeight> & {
```
MattPereira, line 116: This is how to override the generic w/ default!

test/v3/createPool/reClamm/reClamm.integration.test.ts
```diff
@@ -0,0 +1,183 @@
...
+                },
+            ],
+            swapFeePercentage: parseUnits('0.01', 18),
+            pauseManager: zeroAddress,
+            swapFeeManager: zeroAddress,
+            initialMinPrice: parseUnits('0.5', 18),
+            initialMaxPrice: parseUnits('8', 18),
+            initialTargetPrice: parseUnits('3', 18),
+            priceShiftDailyRate: parseUnits('1', 18),
+            centerednessMargin: parseUnits('0.2', 18),
```
MattPereira, line 106: Vaules copied from SC team test

https://github.com/balancer/reclamm/blob/a53760204d76a3f32b2dcb2bfa5f3c89cb67b0ba/test/reClammPool.test.ts#L37-L51

src/utils/numbers.ts
MattPereira, line 1: ### Options
- recreate logic on our own? do we already have this in SDK `MathSol` ?
- maybe this can be installed as npm dependency somehow? ( comes from `v3-monorepo/pvt/helpers` )

src/utils/calculateReClammInitAmounts.ts
```diff
@@ -0,0 +1,59 @@
...
+}): Promise<InputAmount[]> {
+    // fetch proportion value from pool contract
+    const proportion = await client.readContract({
+        address: poolState.address,
+        abi: parseAbi([
+            'function computeInitialBalanceRatio() external view returns (uint256 balanceRatio)',
+        ]),
+        functionName: 'computeInitialBalanceRatio',
+        args: [],
+    });
```
MattPereira, line 28: The way this is currently implemented requires pool to already be deployed, which would wreck the pool creation UX, which has user enter token amounts before deploying the pool

MattPereira: I'm hoping it is possible to recreate logic off chain, otherwise I gotta take a sledge hammer to the creation UX

test/v3/createPool/reClamm/reClamm.integration.test.ts
```diff
@@ -0,0 +1,185 @@
...
```
MattPereira, line 105: Note for Creation UI: 

No matter what the user is going to have to enter these params before choosing a token amount that we will use to calculate the other token amount

src/utils/numbers.ts
johngrantuk: Haven't looked in detail but seems like a lot of this might go away if we assume inputs are always bigint & token scaled (which is what we assume elsewhere)?

brunoguerios: Same here, I didn't look in detail, but assuming `proportion` is fixed point 18 decimals, you should be able to use MathSol for for divDown and mulDown instead of this other function.
I would double check results though, because issues tend to arise when we don't use code that matches exactly the SC implementation.

src/utils/calculateReClammInitAmounts.ts
brunoguerios: If we're already replicating math off-chain on balancer-maths, I'd rather source it from there as well 👍

src/utils/calculateReClammInitAmounts.ts
brunoguerios, line 1: Having in mind that we want the SDK to be a rather lean interface to the SC API, I initially thought it would be best to handle this init amounts calculation on PoolCreator only, while keeping SDK "dumb".
But I like the idea of having this as a helper/utils function 👍 
Similar to what we do with Permit2 stuff, where the SDK itself expects a signature, but we also offer a helper to build it.

test/v3/createPool/reClamm/reClamm.validation.test.ts
brunoguerios, line 1: We usually simply omit `integration` from the file name when splitting into unit/integration tests, but I like the `validation` addition 🙃

brunoguerios (commented):
Nice one 👏 
I like that you opened a draft PR with open discussions that we can keep here within context.
As soon as we follow-up on all of them we should be able to mark the PR as ready for review and approve/merge.

src/utils/calculateReClammInitAmounts.ts
```diff
@@ -0,0 +1,59 @@
...
+    type PublicWalletClient,
+    fpMulDown,
+    fpDivDown,
+    isSameAddress,
+} from 'src';
+import { parseAbi } from 'viem';
+
+export async function calculateReClammInitAmounts({
+    client,
+    poolState,
```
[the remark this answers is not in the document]
MattPereira: could do `tokens` but also need the pool address for the on chain call to `computeInitialBalanceRatio` ( unless we get the maths figured out to replicate off chain )

src/utils/numbers.ts
MattPereira: you were both right thanks for the tips and explanation! :pray:

src/utils/calculateReClammInitAmounts.ts
MattPereira: The maths have been pushed to [johns PR](https://github.com/balancer/balancer-maths/pull/101)

After the release, I will replace this on chain fetch

src/utils/calculateReClammInitAmounts.ts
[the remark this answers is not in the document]
MattPereira: was using `poolstate.address` for the on chain fetch, but will update this to `tokens` once the maths are released

src/utils/calculateReClammInitAmounts.ts
MattPereira: ya also would have been annoying to have to `console.log` the values from some SC test to get the SDK's init ReClamm test to pass :sweat_smile:

test/v3/createPool/reClamm/reClamm.integration.test.ts
```diff
@@ -0,0 +1,246 @@
...
+            rawAmount: parseUnits('69', BAL.decimals),
+            decimals: BAL.decimals,
+        };
+
+        // helper calculates the amount for the other token
+        const amountsIn = await calculateReClammInitAmounts({
+            createPoolInput,
+            tokens: poolState.tokens,
+            givenAmountIn,
+        });
```
MattPereira, line 168: I kinda like asking for fewer inputs for `calculateReClammInitAmounts` helper

but we could also be more explicit if team prefers. this is all that is required from `createPoolInput`

```
   const { initialMinPrice, initialMaxPrice, initialTargetPrice } = createPoolInput;
```

brunoguerios: Since you're already using an object as input, it's quite a low effort to adapt it to be more explicit, right?
If you want to leverage the already existing `createPoolInput`, all you'd have to do is to pass it spreaded (i.e. `...createPoolInput`.
But definitely not a blocker in case you'd like to keep as is 👍

src/utils/calculateReClammInitAmounts.ts
```diff
@@ -0,0 +1,66 @@
...
+    tokens,
+    givenAmountIn,
+    createPoolInput,
+}: {
+    tokens: MinimalToken[];
+    givenAmountIn: InputAmount;
+    createPoolInput: CreatePoolReClammInput;
+}): Promise<InputAmount[]> {
+    // sort the user provided tokens
+    const sortedTokens = tokens.sort((a, b) =>
```
johngrantuk, line 20: We normally sort by token `index`. Is there any reason not to do that this time?

test/v3/createPool/reClamm/reClamm.integration.test.ts
johngrantuk: I prefer more explicit as its easier to follow/debug and takes less work to construct the input.

src/utils/calculateReClammInitAmounts.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    let calculatedAmountIn: InputAmount;
+
+    // https://github.com/balancer/reclamm/blob/8207b33c1ab76de3c42b015bab5210a8436376de/test/reClammPool.test.ts#L120-L128
+    if (givenTokenIndex === 0) {
+        // if chosen token is first in sort order, we multiply
+        calculatedAmountIn = {
+            address: sortedTokens[1].address,
+            rawAmount: MathSol.mulDownFixed(
+                givenAmountIn.rawAmount,
```
johngrantuk, line 45: Haven't had time to confirm the maths but is using `rawAmount` correct here? @brunoguerios can maybe double check during his review.
Also I think its worth adding an integration test that includes a token of <18 decimals. WETH/USDC might be good as there are already some on Base you could use to sanity check values, etc.

src/utils/calculateReClammInitAmounts.ts
```diff
@@ -1,44 +1,46 @@
...
-        address: poolState.address,
-        abi: parseAbi([
-            'function computeInitialBalanceRatio() external view returns (uint256 balanceRatio)',
-        ]),
-        functionName: 'computeInitialBalanceRatio',
-        args: [],
-    });
+    // sort the user provided tokens
+    const sortedTokens = tokens.sort((a, b) =>
+        a.address.localeCompare(b.address),
```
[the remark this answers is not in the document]
brunoguerios: Maybe another way to ask this is: are pool tokens necessarily ordered alphabetically? (that's the assumption here, right?)

src/utils/calculateReClammInitAmounts.ts
brunoguerios: Since these are the amounts to be used in the init, I'd expect it to be rawAmount as well 👍 
But I agree that an integration test is good to rule out any unexpected issues with <18 decimals tokens.

src/utils/calculateReClammInitAmounts.ts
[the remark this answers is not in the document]
MattPereira: I saw the SC team using `.localeCompare` to sort tokens in the reclamm repo

johngrantuk (approved):
As this should have no knock-on effects elsewhere I'll approve in advance of final changes so you can release when you're ready and unblock pool creator.

src/utils/calculateReClammInitAmounts.ts
MattPereira: not sure what alternative to using `rawAmount` would be? :thinking:

MattPereira: not sure what alternative to using `rawAmount` would be? :thinking:

MattPereira: good thinking about adding test that uses token <18 decimals! :raised_hands:

src/utils/calculateReClammInitAmounts.ts
MattPereira: nah i'll make the change to use the SDK's `getSortedTokens` :+1:

src/utils/calculateReClammInitAmounts.ts
brunoguerios: Nice catch.
I think this will be an issue when we have tokens with different decimals as it will keep the decimals of the given token.
I'd suggest casting to `TokenAmount` and using scaled18 for the math. Then pass rawAmount on the result. If should be able to handle scaling up/down for you.
Wondering how rates will affect this 🤔 
Definitely worth having an integration test to make sure it's working as expected 👍

MattPereira: so tests definitely fail when one of the tokens is <18 decimals. 

struggling to understand if it is because of my updated `calculateReClammInitAmounts` logic with the scaling

or if it is because something needs to change with `initialMinPrice`, `initialMaxPrice`, and `initialTargetPrice` if one token has 18 decimals and the other has 6

src/utils/calculateReClammInitAmounts.ts
```diff
@@ -36,25 +38,48 @@ export async function calculateReClammInitAmounts({
...
+        );
         calculatedAmountIn = {
             address: sortedTokens[0].address,
-            rawAmount: MathSol.divDownFixed(
-                givenAmountIn.rawAmount,
-                proportion,
-            ),
+            rawAmount: rawAmountScaledNative.amount,
             decimals: sortedTokens[0].decimals,
         };
```
MattPereira, line 84: request for sanity check :pray:

src/utils/calculateReClammInitAmounts.ts
```diff
@@ -0,0 +1,91 @@
...
+    // https://github.com/balancer/reclamm/blob/8207b33c1ab76de3c42b015bab5210a8436376de/test/reClammPool.test.ts#L120-L128
+    if (givenTokenIndex === 0) {
+        // if given token is first in sort order, we multiply (math must be done in scaled 18)
+        const rawAmountScaled18 = MathSol.mulDownFixed(
+            givenTokenAmountScaled18,
+            proportion,
+        );
+        // then convert back to token decimals raw amount (rounding down)
+        const rawAmountScaledNative = TokenAmount.fromScale18Amount(
+            givenToken,
```
brunoguerios, line 61: Found one issue.
This should be the `calculatedToken` instead of `givenToken`.
Same goes when givenTokenIndex is 1.
I won't be able to apply the fix to validate, but you can try making the change to see if tests pass.

MattPereira: there-goes-my-hero-foo-fighters

PR #660, opened 2025-04-24
Security - Bump dependencies to latest version

### LINKED ISSUE #175: Dependency Opsec Audit/Standards
Following the Ledger issue its worth reviewing. Couple of useful threads:
https://twitter.com/ricmoo/status/1735414433956999177
https://twitter.com/LefterisJP/status/1735389403357319253
https://twitter.com/evan_van_ness/status/1740491595152318623

brunoguerios:
Closes #175 

This PR applies several updates focused on improving the repo security:
- update pnpm to v10.8.1
   - v10 is allegedly safer and has better performance - it introduced `pnpm approve-builds` which should be used to flag which dependency is able to run post install scripts (a common vector of security issues on dependencies)
   - they suggest devs use corepack to manage different versions of pnpm across different projects/repos
   - I had to delete other pnpm installations to make sure it was working properly (e.g. homebrew, npm, etc..)
- update biome to v1.9.4
   - in order to conform to it `biome.json` was updated to use v1.9.4 schema. This file is used by cursor/vscode biome extension to highlight errors/warnings. In case it't no picked up by default, restarting the IDE solved it for me.
   - add `noExplicitAny` as linter rule suggested as security measure (and refactored some of the code to conform to that)
- update all other deps to latest and made semver more strict (i.e. locked versions)
   - should not affect dev exp., but please make sure you run `pnpm install` to get everything up to date
   - from now on, every dependency update will be flagged within PRs and should be reviewed/approved


johngrantuk (approved):
Nice one Bruno! Couple of last confirmations before release:
* Should it be a major release?
* Can we confirm with FE that no issues? (Maybe you've tried running locally?)

brunoguerios:
> * Should it be a major release?
> * Can we confirm with FE that no issues? (Maybe you've tried running locally?)

I chose minor because it does not add any breaking changes to consumers, but not a super strong opinion. 
I'll set it up locally to make sure it indeed does not introduce any issues to the FE.
If it causes issues, I'll bump a major and report to the FE team to fix them.
If not, I'd say it's ok to leave as a minor release.
What do you think?

johngrantuk:
> > * Should it be a major release?
> > * Can we confirm with FE that no issues? (Maybe you've tried running locally?)
> 
> I chose minor because it does not add any breaking changes to consumers, but not a super strong opinion. I'll set it up locally to make sure it indeed does not introduce any issues to the FE. If it causes issues, I'll bump a major and report to the FE team to fix them. If not, I'd say it's ok to leave as a minor release. What do you think?

Yep, makes sense!

mkflow27 (approved):
quite an update for dependencies. Looking forward to seeing how this works out locally with corepack etc.

MattPereira (approved):
Love this! :tada: 

Seems to have fixed annoying biome pop up warning i was always dismissing :raised_hands: 

Abolishing the caret ranges for exact pins is very wise given the stakes :100:

PR #662, opened 2025-04-25
update to reclamm deployment 3

### LINKED ISSUE #661: Update reClamm
New deployments with some changes: https://github.com/balancer/balancer-deployments/pull/244/files

MattPereira:
Closes #661 

### Summary
- minor organization of update deployments script
- updated reclamm addresses and abis to deployment 3
- query init amounts with reclamm pool's `computeInitialBalancesRaw`


src/utils/calculateReClammInitAmounts.ts
```diff
@@ -1,80 +1,94 @@
...
-        rawAmountScaled18,
-    );
+	// Convert back to native token decimals of the calculated token (rounding down)
+	const calculatedTokenScaledNative = calculatedToken.rate
+		? TokenAmountWithRate.fromScale18AmountWithRate(
+				calculatedToken,
+				rawAmountScaled18,
+				calculatedToken.rate,
+		  )
+		: TokenAmount.fromScale18Amount(calculatedToken, rawAmountScaled18);
```
MattPereira, line 82: For the failling test `'6 decimal token with rate given in'`

The amount being calculated for WETH is too large. maybe this is spot that needs fixing. 

played around with only using `TokenAmountWithRate` and passing a default value for rate. this experiment fixes the one failing test but breaks others 

```
	const calculatedTokenScaledNative =
		TokenAmountWithRate.fromScale18AmountWithRate(
			calculatedToken,
			rawAmountScaled18,
			calculatedToken.rate ?? WAD * WAD,
		);
```

src/utils/computeReClammInitAmounts.ts
```diff
@@ -0,0 +1,125 @@
...
+        ? // TODO: this is broken, how to fix so tokens scaled 18 are scaled down by rate?
+          TokenAmountWithRate.fromScale18AmountWithRate(
+              computedToken,
+              computedAmountScaled18,
+              computedToken.rate,
+          )
+        : TokenAmount.fromScale18Amount(computedToken, computedAmountScaled18);
+
+    console.log('computedToken.rate', computedToken.rate);
+    console.log('computedAmountScaledRaw', computedAmountScaledRaw.amount);
```
MattPereira, line 100: `computedAmountScaled18` seems to be fine ( see `initialBalances` )

But I don't yet understand how to scale down by rate :upside_down_face: 

image

src/types.ts
```diff
@@ -1,5 +1,6 @@
 export type Address = `0x${string}`;
 export type Hex = `0x${string}`;
+export type BigintIsh = bigint | string | number;
```
MattPereira, line 3: moved `BigintIsh` type here since it's usage is spread across a number of files 

 ( originally was declared in `tokenAmount.ts` )

src/utils/computeReClammInitAmounts.ts
```diff
@@ -0,0 +1,109 @@
...
+        undefined, // wrapped
+        sortedTokens[calculatedTokenIndex]?.rate,
+    );
+
+    // Scale given amount to 18 decimals before doing math
+    const referenceTokenAmount = referenceToken.rate
+        ? TokenAmountWithRate.fromRawAmountWithRate(
+              referenceToken,
+              referenceAmountIn.rawAmount,
+              referenceToken.rate,
```
brunoguerios, line 66: I'm trying to wrap my head around the idea of adding rate to `Token` 🤔
I was expecting it would be only added to `TokenAmountWithRate` because that's where we keep concepts like amounts, decimals and rates.
Unless I'm missing something, this is the only place you're using it, right? If yes, you should be able to remove it from here without losing much.
You could replace with something like this:
```suggestion
              sortedTokens[givenTokenIndex]?.rate,
```

What do you think?

src/utils/computeReClammInitAmounts.ts
```diff
@@ -0,0 +1,109 @@
...
+            : MathSol.divDownFixed(referenceTokenAmount.scale18, proportion); // if given token is second in sort order, we divide'
+
+    // Convert back to native token decimals of the calculated token (rounding down)
+    const computedAmountScaledRaw = computedToken.rate
+        ? // TODO: this is broken, how to fix so tokens scaled 18 are scaled down by rate?
+          TokenAmountWithRate.fromScale18AmountWithRate(
+              computedToken,
+              computedAmountScaled18,
+              computedToken.rate,
+          )
```
brunoguerios, line 86: This is no longer broken, right?

src/utils/computeReClammInitAmounts.ts
```diff
@@ -0,0 +1,109 @@
...
+    // // on chain sanity check for computedAmountScaled18
+    // const initialBalances = await client.readContract({
+    //     address: poolAddress,
+    //     abi: parseAbi([
+    //         'function computeInitialBalances(address referenceToken, uint256 referenceAmountIn) view returns (uint256[])',
+    //     ]),
+    //     functionName: 'computeInitialBalances',
+    //     args: [referenceToken.address, referenceTokenAmount.scale18],
+    // });
+    // console.log('on chain initialBalances:', initialBalances);
```
brunoguerios, line 104: I'd say these are useful for debugging, but I'd remove them before merging the PR

src/utils/computeReClammInitAmounts.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    // TODO: sort token amounts within InitPool.buildCallWithPermit2 so this not needed
+    // Return amounts in consistent order based on token addresses
+    return [referenceAmountIn, computedAmountIn];
```
brunoguerios, line 108: If you're not doing this on this PR, please create an issue to keep track of it and delete the comment

brunoguerios (commented):
Nice one! 👏 
The only thing I'd like to discuss a bit is rate being added to Token, other than that LGTM 👍

src/utils/computeReClammInitAmounts.ts
MattPereira: my bad i forgot to remove this comment after i realized assumption was incorrect as tokens still sorted by `InitPool.buildCall` during execution of `InitPool.buildCallWithPermit2`

src/utils/computeReClammInitAmounts.ts
MattPereira: Ya idk why I thought that was a good idea yesterday :thinking: 

I agree and like your suggestion much better

MattPereira: oh wait I remember now . i did it because `getSortedTokens` returns type `Token` w/o rate

https://github.com/balancer/b-sdk/blob/3aff1f0640330bee66f339541840d9a1882d7194/src/entities/utils/getSortedTokens.ts#L7-L10

but there are still other ways to solve w/o changing `Token` i will explore

MattPereira: the tricky part is a smooth way to ask for the rate of the computed token as part of input to `computeReClammInitAmounts` :thinking:

MattPereira: @brunoguerios  how would you feel about using the pool address to on chain fetch the token config to get rate provider address then fetch the rate for each token?

MattPereira: we could also fetch the initial params using the pool address so user only has to provide chainId, pool addy, and reference token / amount in

https://github.com/balancer/reclamm/blob/44b1d60324746aaf9c0ef3cbac6ffa763f134531/contracts/interfaces/IReClammPool.sol#L401

src/utils/computeReClammInitAmounts.ts
```diff
@@ -0,0 +1,95 @@
...
+export type ComputeReClammInitAmountInput = {
+    chainId: number;
+    tokens: MinimalTokenWithRate[];
+    referenceAmountIn: InputAmount;
+    initialMinPrice: bigint;
+    initialMaxPrice: bigint;
+    initialTargetPrice: bigint;
+};
+
+export async function computeReClammInitAmounts({
```
mkflow27, line 21: What is the reason this is async? I don't think it needs to be.

src/utils/computeReClammInitAmounts.ts
brunoguerios: Taking a step back, why aren't we querying that onchain? ReClamm has a function to `computeInicialBalances`, right?

src/utils/computeReClammInitAmounts.ts
MattPereira: nice catch! ya didn't need to be but will need after we start making `computeInitialBalances` fetch to pool contract

test/v3/createPool/reClamm/reClamm.integration.test.ts
```diff
@@ -9,6 +9,9 @@ import {
     parseUnits,
     TestActions,
     Hex,
+    parseAbi,
+    erc4626Abi,
+    erc20Abi,
```
MattPereira, line 14: looks like these are not used anymore and can be removed

also theres a commented out import just below this that could be cleaned up

```
    // balancerV3Contracts,
``

brunoguerios (approved):
LGTM ✅

PR #679, opened 2025-05-09
API client headers

gmbronco:
The intention of this PR is to add API client headers for usage tracking. Our partners use the SDK to call the API, and we need to identify their traffic to improve issue resolution and performance optimisation needs.


src/data/providers/balancer-api/client/index.ts
```diff
@@ -3,9 +3,15 @@ import { ChainId } from '../../../../utils';
...
     chainId: ChainId;
+    clientName: string;
+    clientVersion: string;
+
     constructor(apiUrl: string, chainId: ChainId) {
         this.apiUrl = apiUrl;
         this.chainId = chainId;
+
+        this.clientName = process.env.API_CLIENT_NAME || 'balancer-sdk';
+        this.clientVersion = process.env.API_CLIENT_VERSION || __PACKAGE_VERSION__;
```
brunoguerios, line 14: Test is failing because __PACKAGE_VERSION__ is not defined during tests (only during build).
I see that you're assigning JSON.stringify(pkg.version) to it on the build config.
Why don't you simply default the clientVersion here to that, instead of assigning it on build config only?
Here's my suggestion:
```suggestion
        this.clientVersion = process.env.API_CLIENT_VERSION || JSON.stringify(pkg.version);
```

src/globals.d.ts
```diff
@@ -0,0 +1 @@
+declare const __PACKAGE_VERSION__: string;
```
brunoguerios, line 1: Linter fails here. It requires a new line to be added at the end of the file.
```suggestion
declare const __PACKAGE_VERSION__: string;

```

brunoguerios (changes requested):
Should be ok with the fixes I suggested. Let me know what you think.

gmbronco:
All the tests are now passing. Here's a summary of the changes:

1. Updated the `BalancerApiClient` class to:
   - Accept an optional `options` parameter with `clientName` and `clientVersion`
   - Handle both browser and server environments safely
   - Use environment variable names prefixed with `BALANCER_SDK_` for better namespacing
   - Handle the case when `__PACKAGE_VERSION__` is not defined in tests
   - Add proper TypeScript interfaces and documentation

2. Updated the `BalancerApi` class to:
   - Pass the client options to the `BalancerApiClient` constructor
   - Use the `BalancerApiClientOptions` interface

3. Added tests to verify:
   - Default values work correctly
   - Explicit options work correctly
   - Environment variables work correctly
   - Explicit options correctly override environment variables

4. Created a test helper to set `__PACKAGE_VERSION__` in the test environment

With these changes we will be able to set API headers easily in any scenario. The `Version` is adding some boilerplate to the general setup, so let me know if that's acceptable, or needs to be dropped.

brunoguerios (approved):
LGTM ✅

PR #680, opened 2025-05-14
feat: address validation

### LINKED ISSUE #635: Improve input validation for missing configs (for new chains)
Input validation currently does not provide a good error message for missing configs, which usually happens on chains that are yet not supported.

Context/Thoughts
- We currently check if chain exists within chains list, but this is not a good enough check as is. 
- Do we need to enforce configs properly set for every chain in the chains list?
- Should we be more explicit on which chains are supported? (e.g. on docs)
- We should keep in mind that not everything is supported on all chains (e.g.  nested operations are not supported for Balancer v3 on AVAX)

mkflow27:
Closes #635 

This pr adds validation if a given address exists (read: we are deployed there). The validation takes into account a contract name, chainId, and protocolVersion.

After the initial review on the onsite, I have adapted the class based design for address validation.


MattPereira (approved):
lgtm well done ser :tada:

johngrantuk (approved):
Nicely done.

brunoguerios (approved):

PR #686, opened 2025-06-05
feat: add pool abis to decode pool specific errors

### LINKED ISSUE #653: Add LBP ABI to the ABIs used during query
The contract function "querySwapSingleTokenExactIn" reverted with the following signature:
0xfdf79845

Unable to decode signature "0xfdf79845" as it was not found on the provided ABI.
Make sure you are using the correct ABI and that the error exists on it.
You can look up the decoded signature here: https://openchain.xyz/signatures?query=0xfdf79845.
 
Contract Call:
  address:   0x0BF61f706105EA44694f2e92986bD01C39930280
  function:  querySwapSingleTokenExactIn(address pool, address tokenIn, address tokenOut, uint256 exactAmountIn, address sender, bytes userData)
  args:                                 (0x8E4Bd90c02b612594C96a31fADD7E4957bd5daBb, 0xb19382073c7A0aDdbb56Ac6AF1808Fa49e377B75, 0xB77EB1A70A96fDAAeB31DB1b42F2b8b5846b2613, 10000000, 0x0000000000000000000000000000000000000000, 0x)

During test data generation for the SOR I wanted to generate test-data while the sale has not started. The tx reverts but unable to decode the error signature because it is missing from the abis shared with the Swap.

mkflow27:
closes #653

### DIFF
--- a/.changeset/small-llamas-learn.md
+++ b/.changeset/small-llamas-learn.md
@@ -1,0 +1,5 @@
+---
+"@balancer/sdk": minor
+---
+
+additional abis for improved error reporting


test/v3/utils/queryErrorHandling.integration.test.ts
```diff
@@ -0,0 +1,99 @@
...
+        })
+            .extend(publicActions)
+            .extend(walletActions);
+
+        testAddress = (await client.getAddresses())[0];
+    });
+
+    beforeEach(() => {
+        // nothing needed
+    });
```
brunoguerios, line 62: You should be able to remove this if it's not needed

brunoguerios (approved):
Nice one! LGTM ✅ 
Ps: some that are easily reproducible are `MaxInRatio` and `MaxInvariantRatio` - would be nice to have tests covering them as well

PR #696, opened 2025-07-03
feat: add lbpWithMigration & migratePool

mkflow27:
This pr adds:
- Creating lbps with the intention of migratinv liquidity and without migrating liquidity
- sdk support for migrating liquidity

Since we seem to be having issues with initializing pool with latest block numbers tests are not included but example scripts are given. 

A new entity called `MigratePool` is created utilising a factory pattern (we use this in other entities as well, such as `CreatePool`). This is created to have the option for further extensibility if it is ever required. 

Outstanding:
- Update pauseManager, swapFeeManager, etc to follow other pool Creation conventions.


src/entities/migratePool/types.ts
```diff
@@ -0,0 +1,82 @@
...
+    buildCall(input: MigratePoolInput): MigratePoolBuildCallOutput;
+}
+
+export type MigratePoolInput = {
+    poolType: string;
+    pool: Address;
+    chainid: number;
+    rpcUrl: string;
+    excessReceiver: Address;
+    weightedPoolParams: WeightedPoolParams;
```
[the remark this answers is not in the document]
mkflow27: Yep, these params are used to create the pool

src/entities/migratePool/index.ts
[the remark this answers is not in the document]
brunoguerios: This addition is still pending, right?

brunoguerios (changes requested):
Seems like some tests are failing. Is that expected?

src/entities/createPool/types.ts
```diff
@@ -180,4 +181,17 @@ export type CreatePoolLiquidityBootstrappingInput = Omit<
 > & {
     lbpParams: LBPParams;
     poolType: PoolType.LiquidityBootstrapping;
+    poolCreator?: Address;
 };
+
+export type LBPMigrationParams = {
+    bptLockDuration: bigint;
```
johngrantuk, line 188: Would be useful to change to something that describes time unit? e.g. `bptLockDurationMinutes`

mkflow27: I have changed the lock duration to now be `bptLockDurationinSeconds`. About start time and endtime I could see it be changed to `startTimestamp` and `endTimestamp`

mkflow27 (commented):
> Seems like some tests are failing. Is that expected? 

https://github.com/balancer/b-sdk/pull/696#discussion_r2287485734. The failing lbp tests is due to the conflict of either using the current migrationFactory or a working blockNumber (where the factory is deployed)

src/entities/migratePool/index.ts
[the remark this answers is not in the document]
brunoguerios: Yes, I'm referring to the constructor. The implementation should follow a factory pattern (similar do AddLiquidity) where an SDK consumer can feed/inject custom types instead of having to rely on existing/default ones. This provides extensibility to the SDK.

src/entities/migratePool/liquidityBootstrapping/index.ts
```diff
@@ -0,0 +1,124 @@
...
+        return {
+            callData,
+            to: AddressProvider.LBPoolMigrationRouter(input.chainid),
+        };
+    }
+    public buildCallWithPermit2(
+        _input: MigratePoolLiquidityBootstrappingInput,
+        _permit2: Permit2,
+    ): MigratePoolBuildCallOutput {
+        // not implemented
```
brunoguerios, line 87: Any reason why this wasn't implemented? If the idea is to avoid blocking this PR from being released, I'd say it would be nice to create a ticket on the project board to handle this in the future.

mkflow27: I have now removed that functionality from the `MigratePoolBase` interface. The Migration Router does not support Permit in its current iteration.

PR #726, opened 2025-09-29
Fix scientific notation parsing for slippage inputs

brunoguerios:


MattPereira (approved):
Well done and such thorough testing! 👌

PR #747, opened 2025-12-04
Add Liquidity Unbalanced Via Swap

mkflow27:
Closes #796 

This PR proposes a solution for adding liquidity single sided to 2 token pools using the AddLiquidityUnbalancedViaSwapRouter.

The main problem to solve is that we want to allow users to simply input amounts in, but the router operation also expects an `exactBptAmountOut`.

In order to solve that, we'll need to calculate/estimate the exactBptAmountOut corresponding to the provided amounts in.

The implemented approach relies on an initial guess which is then adjusted/corrected iteratively based on how far they are from on-chain query results. A full description of each step can be found documented both in the codebase and also in this [Notion doc (with illustrative example)](https://www.notion.so/Add-Liquidity-Unbalanced-via-Swap-Algorithm-Explanation-2f6e395e228180509d99f560d1ab2246?source=copy_link).

Integration tests have shown the approach is able to calculate/estimate BPT amounts that approximate the amounts in provided within a tolerance of 0.01% even for add liquidity amounts up to 60% of the balance of the pool.


test/v3/addLiquidityUnbalancedViaSwap/addLiquidityUnbalancedViaSwap.integration.test.ts
johngrantuk: Can we remove the Gnosis test in favour of having one integration file?

test/v3/addLiquidityUnbalancedViaSwap/addLiquidityUnbalancedViaSwap.test.ts
```diff
@@ -0,0 +1,420 @@
...
+                    ...mockQueryOutput,
+                    slippage: Slippage.fromPercentage('1'),
+                    deadline: maxUint256,
+                };
+
+                const result =
+                    addLiquidityUnbalancedViaSwap.buildCall(buildCallInput);
+
+                expect(result).toHaveProperty('callData');
+                expect(result).toHaveProperty('to');
```
johngrantuk: Should be able to test this is known contract address? Possible same with some other values?

src/entities/addLiquidityUnbalancedViaSwap/index.ts
```diff
@@ -0,0 +1,212 @@
+import { encodeFunctionData, maxUint256, parseEther, zeroAddress } from 'viem';
```
johngrantuk: Noticed theres quite a few unused imports that don't seem to be picked up.

src/entities/addLiquidityUnbalancedViaSwap/index.ts
```diff
@@ -0,0 +1,212 @@
...
+            );
+        }
+
+        const bptToken = new Token(input.chainId, poolState.address, 18);
+        const bptOut = TokenAmount.fromRawAmount(bptToken, currentBptAmount);
+
+        const output: AddLiquidityUnbalancedViaSwapQueryOutput = {
+            pool: poolState.address,
+            bptOut,
+            exactAmountIn: TokenAmount.fromRawAmount(
```
johngrantuk: Does this need to be returned? Might be confusing to user. Instead we can just hardcode 0 amount in buildCall?

PR #823, opened 2026-04-22
Fix boosted V3 proportional remove: per-index tokensOut inference

johngrantuk:
RemoveLiquidityBoostedV3.query previously inferred unwrapWrapped by building a flat address → metadata map over all pool tokens and underlyings, then sorting the resolved entries by index. This fails in two related ways when a pool contains a plain token whose address is also the underlying of an ERC4626 slot (e.g. a ZCHF + svZCHF stable pool):

The flat map overwrites the plain-token entry with the ERC4626-underlying entry for the same address, so both slots resolve as isUnderlyingToken: true → unwrapWrapped: [true, true].
The amountsOut decimals lookup used the same flat map, meaning both entries picked up the metadata for the wrong slot.
Together these produced unwrapWrapped: [true, true] and an on-chain BufferNotInitialized revert for the non-ERC4626 leg.

### Problem
`RemoveLiquidityBoostedV3.query` previously inferred `unwrapWrapped` by building a flat `address → metadata` map over all pool tokens and underlyings, then sorting the resolved entries by index. This fails when a pool contains a plain token whose address is also the underlying of an ERC4626 slot (e.g. a ZCHF + svZCHF stable pool):
1. The flat map overwrites the plain-token entry with the ERC4626-underlying entry for the same address, so both slots resolve as `isUnderlyingToken: true` → `unwrapWrapped: [true, true]`.
2. The `amountsOut` decimals lookup used the same flat map, so both entries picked up the metadata for the wrong slot.
Together these produced `unwrapWrapped: [true, true]` and an on-chain `BufferNotInitialized` revert for the non-ERC4626 leg.
### Changes
**`src/entities/removeLiquidityBoosted/inferUnwrapWrapped.ts`** *(new)*
Pure helper that resolves `unwrapWrapped[i]` per pool slot: for each `i`, `tokensOut[i]` is compared to `poolTokens[i].address` (→ `false`) and, if the slot has an underlying, to `poolTokens[i].underlyingToken.address` (→ `true`). Any other address throws a clear validation error referencing the exact slot index.
**`src/entities/removeLiquidityBoosted/index.ts`**
- Replaced the old map+sort `unwrapWrapped` block with a call to `inferUnwrapWrapped`.
- Fixed `amountsOut` construction to derive decimals from `poolTokens[i]` directly (using `underlyingToken.decimals` when unwrapping, `pt.decimals` otherwise) instead of looking up the flat map.
- Removed the now-unused `buildPoolStateTokenMap` import.
**`src/entities/removeLiquidityBoosted/types.ts`**
Added JSDoc on `tokensOut` documenting the pool-index-order contract, the three valid cases (pool token / ERC4626 underlying / validation error), and the reason the ordering requirement exists.
**`src/entities/priceImpact/addLiquidityUnbalancedBoosted.ts`**
Added an inline comment on the `tokensOut: poolTokens.map(t => t.address)` line noting that `poolTokens` is already sorted by index (via `getSortedTokens`), so it satisfies the new contract without modification.
**`test/v3/removeLiquidityBoosted/inferUnwrapWrapped.test.ts`**
Activated all 8 unit assertions (previously `it.todo` stubs), covering: duplicate-address pool (ZCHF + svZCHF), standard pool, validation errors with correct slot-index references, and case-insensitive address matching.
**`test/v3/removeLiquidityBoosted/removeLiquidityPartialBoosted.integration.test.ts`**
Updated `tokensOut` from `[USDT.address, WETH.address]` to `[WETH.address, USDT.address]` (WETH is at pool index 0, stataUSDT/USDT at index 1) and aligned the `validateTokenAmounts` assertion order to match.
### Breaking change
Callers must now pass `tokensOut` in **pool token index order** (matching `poolState.tokens` sorted by `index`). The previous implementation sorted by resolved index after the fact, tolerating any unique-address ordering; that behaviour is removed.
**Example:** `partialBoostedPool_WETH_stataUSDT` has WETH at index 0 and stataUSDT at index 1. Callers previously passing `[USDT, WETH]` must now pass `[WETH, USDT]`.

### DIFF
--- a/src/entities/removeLiquidityBoosted/inferUnwrapWrapped.ts
+++ b/src/entities/removeLiquidityBoosted/inferUnwrapWrapped.ts
@@ -1,0 +1,47 @@
+import { Address } from 'viem';
+
+import { PoolTokenWithUnderlying } from '@/entities/types';
+import { isSameAddress, inputValidationError } from '@/utils';
+
+/**
+ * Derives the per-slot `unwrapWrapped` boolean array for a boosted proportional
+ * remove-liquidity query.
+ *
+ * @param poolTokens - Pool tokens (any order; sorted internally by `index`).
+ * @param tokensOut  - One address per pool slot, aligned with sorted `index`.
+ *                    Each entry must be either the pool token address (→ false)
+ *                    or, for ERC4626 slots, the underlying address (→ true).
+ *
+ * This is a pure function: all validation that is input-shape dependent
+ * (e.g. length mismatch) is handled by the caller via InputValidator before
+ * this helper is invoked.
+ */
+export function inferUnwrapWrapped(
+    poolTokens: PoolTokenWithUnderlying[],
+    tokensOut: Address[],
+): boolean[] {
+    const sorted = [...poolTokens].sort((a, b) => a.index - b.index);
+
+    return sorted.map((pt, i) => {
+        const desired = tokensOut[i];
+
+        if (isSameAddress(desired, pt.address)) {
+            return false;
+        }
+
+        if (
+            pt.underlyingToken !== null &&
+            isSameAddress(desired, pt.underlyingToken.address)
+        ) {
+            return true;
+        }
+
+        const underlyingHint = pt.underlyingToken
+            ? ` or its underlying ${pt.underlyingToken.address}`
+            : '';
+        throw inputValidationError(
+            'Remove Liquidity Boosted',
+            `tokensOut[${i}] (${desired}) is not valid for pool token index ${i}: expected pool token ${pt.address}${underlyingHint}`,
+        );
+    });
+}
--- a/test/v3/removeLiquidityBoosted/inferUnwrapWrapped.test.ts
+++ b/test/v3/removeLiquidityBoosted/inferUnwrapWrapped.test.ts
@@ -43,7 +43,8 @@ const duplicateAddressPool = {
 // a distinct underlying.
 const PLAIN_A = '0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa' as `0x${string}`;
 const ERC4626_B = '0xbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb' as `0x${string}`;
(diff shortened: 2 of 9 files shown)


test/v3/removeLiquidityBoosted/inferUnwrapWrapped.test.ts
```diff
@@ -0,0 +1,141 @@
+// pnpm test -- inferUnwrapWrapped.test.ts
```
MattPereira, line 1: nbd but seems the `--` syntax invalidates the target single test arg and runs them all

src/entities/removeLiquidityBoosted/inferUnwrapWrapped.ts
```diff
@@ -0,0 +1,47 @@
...
+        ) {
+            return true;
+        }
+
+        const underlyingHint = pt.underlyingToken
+            ? ` or its underlying ${pt.underlyingToken.address}`
+            : '';
+        throw inputValidationError(
+            'Remove Liquidity Boosted',
+            `tokensOut[${i}] (${desired}) is not valid for pool token index ${i}: expected pool token ${pt.address}${underlyingHint}`,
```
MattPereira, line 44: 👌

MattPereira (approved):
lgtm requiring sorted `tokensOut` seems more robust & nice docs on the breaking change 🙌

brunoguerios (commented):
Quoting Juani: "why do people do this? 🫠" 😂 
Shouldn't a buffer be enough for swapping between wrapped/underlying - why was this pool created like this?
Anyway, I'd avoid the approach of requiring users to provide a sorted tokensOut array as input if possible.
That breaks the pattern we set on all other SDK inputs, where we allow tokens to be provided in any order to facilitate integration.
I see why poolStateTokenMap won't work here, but there should be away to identify if a token should be unwrapped without having to require sorted inputs.

brunoguerios:
Actually - I remember now that we discussed the scenario of a pool having the same "underlying token" and at the time we reached the conclusion that pools like these wouldn't be supported. That does break the interface because the SDK expects token addresses provided to be unique within the pool. I'd consider pushing back and challenging the need for the pool as a whole.

